Issue #2, opened 2024-10-16
CRR-GridGen Strict Testing


manishvenu:
What the title says, implement rigorous testing for the Grid Gen Module

PR #22, opened 2024-11-05
Start of Docs & Testing for CRR: Driver, GridGen, RCG

### LINKED ISSUE #21: CRR Driver Testing + Documentation

manishvenu:
Fixes #21, #2, https://github.com/CROCODILE-CESM/regional-casegen/issues/8

This PR implements the first test suite and documentation for CRR.

- [ ] Clean & Write up the docstrings for the modules of CRR with good autodoc display
- [ ] Implements a minimal test suite for the driver module. RCG is getting replaced with VCG, and grid_gen with mom6_bathy.
- [ ] Basic Github Workflow for Tests


manishvenu:
Hey @alperaltuntas,

I think this is ready now. I removed the regression testing & added a generic github workflow. One question is that the workflow is set to look in the `tests/` directory, so it doesn't run the RM6 tests as well. I think that's what we want, but if you think it'd be better, we can run the RM6 tests alongside the CrocoDash tests as part of the workflow.

alperaltuntas (approved):

PR #26, opened 2025-01-06
Integrate visualCaseGen into CrocoDash

alperaltuntas:
This PR integrates visualCaseGen into CrocoDash as a backend for CESM-specific features and MOM6 domain generation library. The main change is in the newly added `case` module, which  represents a regional MOM6 case within the CESM framework. It is similar to the `Experiment` class in the regional_mom6 package, but with modifications to work within the CESM framework. This class automates the steps of creating a CESM case and configuring all of its build-time and runtime parameters. Additionally the `Grid`, `Topo` and `VGrid` classes are introduced: These are basically wrappers of corresponding `mom6_bathy` classes (that comes with visualCaseGen). The utilization of these new classes are showcased in the modified `minimal_demo_rect.ipynb` and `minimal_demo_subset_global.ipynb` demo notebooks. The two remaining demo notebooks will need to be transformed subsequently as well.

Note: this PR should be evaluated in conjunction with https://github.com/CROCODILE-CESM/regional-mom6/pull/39

There are several required TODOs:

 - [x] Transform the remaining two notebooks.
 - [ ] Add/adjust tests.
 - [x] Remove modules such as `driver` and `grid_gen` and the entire `regional_casegen` which are no longer necessary.
 - [ ] There are a few datasets (such as the global NWA12 grid and the tidal dataset) that are needed for the demos. We should make them available online, possibly on the svn inputdata repository.)
 - [x] Multi-instance feature.

There are several nice-to-have or longer-term TODOs:

 - use copernicusmarine Python API instead of directing users to CLI. 
 - When CrocoDash is executed on a derecho login node, the kernel crashes when computationally heavy methods are called. We should seek ways of making those more robust or better informing the users about the cause of crashes.
 -  Implement mask_disconnected_ocean_areas() in mom6_bathy, possibly via TopoEditor widget. We should implement a *magic wand*-like  area picker in TopoEditor.
 - Standardize logging across all the libraries.


manishvenu (approved):
Looks good, I'll start work on the required TBD items. We mayneed to change the forcing function calls to regional_mom6 when the PR gets merged, the tidal function arguments changed to just point to the two files & there's additional optional arguments for the forcings (rotation & masking)

alperaltuntas:
@manishvenu Merging the latest main into this PR branch led to CI test failures. From a quick look, it seems the conftest files of CrocoDash and visualCaseGen are conflicting (?), altthough I am not sure why CrocoDash is looking into visualCaseGen conftest file. Any idea of how to fix this?

manishvenu:
> @manishvenu Merging the latest main into this PR branch led to CI test failures. From a quick look, it seems the conftest files of CrocoDash and visualCaseGen are conflicting (?), altthough I am not sure why CrocoDash is looking into visualCaseGen conftest file. Any idea of how to fix this?

Hey @alperaltuntas, it looks like pytest has difficulty importing two files of the same name. I played around with it, but it seems like the easiest way right now is to rename the CrocoDash folder to something different than "tests" (maybe tests_CrocoDash" or take out the __init__.py file in the tests folder.

alperaltuntas:
@manishvenu Any idea why the last commit fails the CI test?

manishvenu:
> @manishvenu Any idea why the last commit fails the CI test?

Hey @alperaltuntas, the conda action fails sometimes I think (the process '/usr/share/miniconda/condabin/conda' failed with exit code 1). Usually works in a few hours or the next day. It's pretty early in the action so it couldn't be much on our side.

manishvenu:
> > @manishvenu Any idea why the last commit fails the CI test?
> 
> Hey @alperaltuntas, the conda action fails sometimes I think (the process '/usr/share/miniconda/condabin/conda' failed with exit code 1). Usually works in a few hours or the next day. It's pretty early in the action so it couldn't be much on our side.
> 
> The only update was mom6_bathy in the last commit, so the CI tests shouldn't change.

I'll really quickly trying creating the environment myself and see what happens

manishvenu:
> > > @manishvenu Any idea why the last commit fails the CI test?
> > 
> > 
> > Hey @alperaltuntas, the conda action fails sometimes I think (the process '/usr/share/miniconda/condabin/conda' failed with exit code 1). Usually works in a few hours or the next day. It's pretty early in the action so it couldn't be much on our side.
> > The only update was mom6_bathy in the last commit, so the CI tests shouldn't change.
> 
> That actually could be the issue, I'll really quickly trying creating the environment myself and see what happens

It is a pip issue, let me investigate

manishvenu:
> > > > @manishvenu Any idea why the last commit fails the CI test?
> > > 
> > > 
> > > Hey @alperaltuntas, the conda action fails sometimes I think (the process '/usr/share/miniconda/condabin/conda' failed with exit code 1). Usually works in a few hours or the next day. It's pretty early in the action so it couldn't be much on our side.
> > > The only update was mom6_bathy in the last commit, so the CI tests shouldn't change.
> > 
> > 
> > That actually could be the issue, I'll really quickly trying creating the environment myself and see what happens
> 
> It is a pip issue, let me investigate
@alperaltuntas It's good now! I didn't correctly revert the merge with main from mom6_bathy so setup.py was the wrong python version.

PR #27, opened 2025-01-15
Tests , Documentation, and Minor Code Edits for the VCG PR

manishvenu:
Tasks:

- [x] Write basic test cases for the new modules in CrocoDash
- [x] Write up some light docs for the new structure
- [x] (Think about) Add some minor convenience params to process_forcings & to case_creation for GLORYS data
- [ ] (Think about) Heavy CESM test case 
- [x] Maybe demo showing how to add grids already generated to CrocoDash

Changes:

- Test cases: Added some test cases for the case module functions and added small connection tests for all other modules (test_minimal_modules)
- Make demos have override=False so new users don't fall in the hole of deleting old work
- Write light doc changes to help new users.
- Add params to process_forcings so users can decide which forcings to run if needed. (This helps mostly with dev work, but also in case the user changes RM6 code for a specific function and wants to rerun the specific forcing function. It also saves a bit of time since initial_condition/tides doesn't need to be called everytime)
- New demo from user grids

TBD:

- [x] Merge the data access PR and write a couple data access PR tests

Isn't Done:

- [ ] Run CESM on github actions in `test_workflow.py::run_full_workflow`


manishvenu:
Two points:

1. I didn't have a great example on how to setup the CESM in github actions to run a small regional case. Maybe table this for later?
3. I merged the data_access branch into this one, which is why it looks like so many files changed

manishvenu:
@alperaltuntas This is ready for review, but I thought I'd table running the CESM in github actions for later unless there's an easy way to do it!

alperaltuntas (approved):

PR #31, opened 2025-02-03
Minor spelling error in case.py process_forcings().

AidanJanney:
**Description:**
There was a bug found when calling `case.process_forcings()` in minimal_demo_rect.ipynb. Process_forcings sets the boundary_type="rectangle" which throws an error when regional_mom6.py function setup_boundary_tides() checks that the boundary_type is either "rectangular" or "curvilinear." 

**Major Changes**
Adjusted the boundary_type assignment in process_forcings() to `boundary_type="rectangular"`

### DIFF
--- a/CrocoDash/case.py
+++ b/CrocoDash/case.py
@@ -388,7 +388,7 @@ class Case:
         if self.tidal_constituents:
 
             if self.ocn_grid.is_rectangular():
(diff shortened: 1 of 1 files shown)


manishvenu (approved):
This is great! Nice catch!

Issue #43, opened 2025-03-14
Integrate large data access workflow with the case module


manishvenu:
Tasks:

- Add a too_much_data bool to trigger the workflow in case.configure_forcings
- With that bool, generates a python script or scripts that calls get_data_piecewise, regrid_piecewise, and merge
- In processing forcings, if too_much_data is triggered, skip regridding of boundary and check that the files exist

PR #48, opened 2025-03-21
Integrate the Large Data Workflow into the Case module 

manishvenu:
Fixes https://github.com/CROCODILE-CESM/CrocoDash/issues/43

Details:
https://github.com/CROCODILE-CESM/CrocoDash/issues/43

Changes:

1. Add forcing varnames into the ProductFunctionRegistry class
2. Create functions to create the large data workflow folder a user can run.
3. Some minor test edits


AidanJanney (approved):
This looks great! Thank you for setting up all of the config and case infrastructure for the data workflow. I'll get started on the get and merge functionality right away!

CrocoDash/case.py
```diff
@@ -2,15 +2,16 @@
...
 import regional_mom6 as rmom6
 from CrocoDash.grid import Grid
 from CrocoDash.topo import Topo
 from CrocoDash.vgrid import VGrid
 from CrocoDash.data_access import driver as dv
-from CrocoDash.data_access import tables as tb
+from CrocoDash.data_access import config as tb
 from CrocoDash.data_access import driver as dv
-from CrocoDash.data_access import tables as tb
+from CrocoDash.data_access import config as tb
```
alperaltuntas, line 14: Why are these import statements duplicated?

CrocoDash/data_access/large_data_workflow/driver.py
```diff
@@ -0,0 +1,80 @@
+import sys
+import json
+from pathlib import Path
+
+sys.path.append("code")
```
alperaltuntas, line 5: Could you explain why this is necessary?

tests/fixtures/state.py
```diff
@@ -6,10 +6,10 @@
 @pytest.fixture(scope="session")
 def is_glade_file_system():
     # Get the hostname
-    hostname = socket.gethostname()
```
alperaltuntas, line 9: A general comment about this module: I think we should implement these features within visualCaseGen, since visualCaseGen has already similar functionalities within its CIME_interface module. We should generalize those functions / code blocks so as to make them useful and be able to replace the functions within this module. That should eliminate the risk of code duplication and be more in line with the DRY principle.

CrocoDash/case.py
manishvenu: Mistake! I'll remove that

CrocoDash/data_access/large_data_workflow/driver.py
manishvenu: If we copy this folder into a provided location, I added that so we can import the scripts in the driver module

tests/fixtures/state.py
manishvenu: Ok, I'll look into this

PR #50, opened 2025-03-26
Create a get_data_piecewise function

### LINKED ISSUE #41: Create a get_data_piecewise function
Create a script to grab forcing data piecewise with a step argument and date specific output file generation. This script will manipulate dates to grab easily processible data access segments with logical defaults. This script will:

- Grab unprocessed forcing data in easily processible segments using a step argument and saving to date specific file names
- Integrate with the data access ProductFunctionRegistry as a class function that can be called and takes advantage of the strict setup of the data access functions.

manishvenu:
Fixes #41 

Details:
#41 

Changes:
A get_data_piecewise script/function in data_access

### DIFF
--- a/CrocoDash/data_access/large_data_workflow/code/get_data_piecewise.py
+++ b/CrocoDash/data_access/large_data_workflow/code/get_data_piecewise.py
@@ -45,6 +45,8 @@ def get_data_piecewise(
         The directory to save the output NetCDF files.
     boundary_number_conversion : dict
         Dictionary mapping boundaries to their numerical identifiers.
+    preview : bool
+        Whether or not to preview the run, default is false
 
     Raises
     ------
--- a/CrocoDash/data_access/large_data_workflow/code/get_data_piecewise.py
+++ b/CrocoDash/data_access/large_data_workflow/code/get_data_piecewise.py
@@ -1,0 +1,110 @@
+from pathlib import Path
+from CrocoDash import utils
+from CrocoDash.data_access.driver import get_rectangular_segment_info
+from CrocoDash.data_access.utils import load_config
+import CrocoDash.data_access.datasets as datasets
+import importlib
+import json
+import re
+import os
+from datetime import datetime
+import xarray as xr
+import pandas as pd
+
+logger = utils.setup_logger(__name__)
+
+def get_data_piecewise(
+    product_name: str,
+    function_name: str,
+    output_dataset_regex: str, # haven't figured out how to process this for output
+    date_format: str,
+    start_date: str,
+    end_date: str,
+    hgrid_path: str | Path,
+    dataset_varnames: dict,
+    step_days: int, # step increment for chunking data
+    output_dir: str | Path,
+    boundary_number_conversion: dict, # doesn't need to be the full dict, but we need to know which boundaries to get
+):
+    
+    ## Get function from datasets module
+    module = importlib.import_module(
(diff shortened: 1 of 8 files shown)


CrocoDash/data_access/large_data_workflow/code/get_data_piecewise.py
```diff
@@ -0,0 +1,166 @@
...
+        pd.date_range(start=start_date, end=end_date, freq=f"{step_days}D")
+        .to_pydatetime()
+        .tolist()
+    )
+
+    # Add the end date manually if not included
+    if dates[-1] != datetime.strptime(start_date, date_format):
+        dates.append(datetime.strptime(end_date, date_format))
+
+    num_files = len(dates) - 1
```
alperaltuntas, line 83: The reason why num_files one less than `len(dates)` is unclear to me, so it should probably be briefly explained in a preceding comment line.

CrocoDash/data_access/large_data_workflow/code/get_data_piecewise.py
```diff
@@ -0,0 +1,166 @@
...
+    -------
+    None
+        Saves the retrieved data to the specified output directory.
+    """
+
+    ## Initialize PFD
+    ProductFunctionRegistry = dv.ProductFunctionRegistry()
+    ProductFunctionRegistry.load_functions()
+    ProductFunctionRegistry.validate_function(product_name, function_name)
+    func = ProductFunctionRegistry.functions[product_name][function_name]
```
alperaltuntas, line 66: Maybe a bit more descriptive alias, instead of just `func`?

CrocoDash/data_access/large_data_workflow/driver.py
```diff
@@ -12,6 +12,10 @@ def main():
     with open(config_path, "r") as f:
         config = json.load(f)
```
alperaltuntas, line 13: Do we have a class to encapsulate the config json files? If not, I think we should introduce one, instead of relying on just `json.load`. That way, we can put in place schema checks to make sure that all necessary information is present when reading or writing these json files. For instance, a few lines below, you check `config["params"]["step"] <= 0`, but is `config["params"]["step"]`  guaranteed to be present in the config json file? A schema check right after json.load would catch any violation. (Both the json.load and the schema check could be carried out in __init__ method of a Config class).

CrocoDash/data_access/large_data_workflow/code/get_data_piecewise.py
```diff
@@ -0,0 +1,166 @@
...
+
+    # Set up date range
+    dates = (
+        pd.date_range(start=start_date, end=end_date, freq=f"{step_days}D")
+        .to_pydatetime()
+        .tolist()
+    )
+
+    # Add the end date manually if not included
+    if dates[-1] != datetime.strptime(start_date, date_format):
```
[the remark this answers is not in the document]
manishvenu: Yup, thanks!

CrocoDash/data_access/large_data_workflow/driver.py
manishvenu: We don't! I can do that! Thanks

manishvenu: Actually, if it's not a huge issue here, I'll do it in the final PR from large_data_workflow to main. All the scripts use a config file, so it would be a bit simpler.

PR #51, opened 2025-03-28
Create the merge_piecewise_dataset script

### LINKED ISSUE #42: Create merge_dataset_piecewise function
After regridding the piecewise data for boundary conditions, we need to merge all of the individual files together – in an efficient way. 

- Make sure dask is installed in default directory _(it is, dask comes with xarray, and is default with `xr.open_mfdataset`)_
- Merge piecewise data, with option to specify dates (or just merge all available boundary files for a given date)
- Be able to specify which boundaries you want to merge

manishvenu:
Fixes #42 

Details:
#42 

Changes:

1. Add a merge_dataset_piecewise script that reads from config
2. Add fixture for formatted OBC
3. Move parse_dataset to utils

### DIFF
--- a/CrocoDash/data_access/large_data_workflow/code/merge_piecewise_dataset.py
+++ b/CrocoDash/data_access/large_data_workflow/code/merge_piecewise_dataset.py
@@ -19,8 +19,40 @@ def merge_piecewise_dataset(
     end_date: str,
     boundary_number_conversion: dict,
     output_folder: str | Path,
+    preview: bool = False,
 ):
-    """ """
+    """
+    Merges piecewise datasets from a folder into consolidated NetCDF files by boundary.
+
+    Parameters
+    ----------
+    folder : str or Path
+        Path to the folder containing the regridded dataset files.
+    input_dataset_regex : str
+        Regular expression pattern to match dataset files.
+    date_format : str
+        Date format string used for parsing the dataset filenames.
+    start_date : str
+        Start date in the specified format.
+    end_date : str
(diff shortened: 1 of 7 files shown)


CrocoDash/data_access/large_data_workflow/code/merge_piecewise_dataset.py
```diff
@@ -0,0 +1,119 @@
...
+        config["boundary_number_conversion"],
+        config["paths"]["merged_dataset_path"],
+    )
+    return
+
+
+if __name__ == "__main__":
+    main(
+        "/glade/u/home/manishrv/documents/croc/regional_mom_workflows/CrocoDash/CrocoDash/data_access/large_data_workflow/config.json"
+    )
```
alperaltuntas, line 119: Is this path string left out by mistake?

alperaltuntas: Oh, I see now that it's an argument to the main function. It wasn't clear from the split view. Regardless, can we just add config.json to CrocoDash, either directly or using something like `pooch` to make this generalizable to other machines. We should avoid hard-coded /glade/ paths. In fact, we should probably add a CI check on that.

manishvenu: Yes! Sorry about this, because this a draft PR before the final merge I just left it in, I'll remove it in the main large_data_workflow to main PR, and add a quick CI check as well.

PR #62, opened 2025-05-07
Fix issue where CrocoDash cases couldn't be called on Casper

### LINKED ISSUE #38: Creating a derecho case using casper fails with machine error in the last step of the Case Constructor
Creating a derecho case using casper fails with machine error in the last step of the Case Constructor. The connection with the CIME object seems to need something additional to let it know that we want the case on derecho.

Failed Step:
`self._cime_case = self.cime.get_case(self.caseroot)`

manishvenu:
Closes #38 

Fix issue where CrocoDash cases couldn't be called on Casper by making the VisualCaseGen CaseCreator object a object on the case and then using the `_is_non_local` function in the following call to CIME.

### DIFF
--- a/CrocoDash/case.py
+++ b/CrocoDash/case.py
@@ -127,7 +127,7 @@ class Case:
 
         self._create_newcase()
 
-        self._cime_case = self.cime.get_case(self.caseroot)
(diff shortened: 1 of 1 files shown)


alperaltuntas (approved):
Awesome!

PR #73, opened 2025-06-10
Large Bathymetry Regrid

AidanJanney:
Refactored regional-mom6 bathymetry regridding to break apart the process for an easier workflow when regridding with mpi. Still need a lot of documentation updates, doc strings, and consultation with Manish on structure, workflow, and formatting.


AidanJanney:
Currently point to a gist with an example PBS script, consider auto generating a PBS, slurm, and bash script in that directory. Clearly labeled. (could have account and info like that as optional args?)

manishvenu:
I believe this will be closed because it's partly done here https://github.com/NCAR/mom6_bathy/pull/34 (with input from Aidan)! Please open a new PR if you have any remaining changes when this is merged Aidan!

Issue #61, opened 2025-04-29
Compset in Case.py doesn't do anything


manishvenu:
The Case.py self.compset isn't used by VisualCaseGen case_creator to create the case. This needs to be fixed! 

Workaround: 

Before self._run_create_newcase(caseroot, compset, resolution, do_exec) in VCG->custom_widget_types->case_creator.py add the compset as a string

PR #78, opened 2025-06-13
Enable compset specification for Case instances

alperaltuntas:
This PR enables compset specification for Case instances, and introduces a G_JRA example (w/ MOM6, CICE and DATM%JRA).

Changes:
- Update visualCaseGen (and mom6_bathy) submodule to enable quick subsampling and CICE compsets.
- Add `compset` argument to Case constructor, which is taken into account for when visualCaseGen `cvars`  are assigned values.
- Update tests and demos.
- Add a MOM6+CICE demo.(This is a very preliminary test case that ran okay for 5 days. We should improve it down the road.)
-
Fixes: #61


manishvenu:
@silverfoxpt 

Hey Hung,

Can you revert your commit and readd it as a PR into this branch? We will need to hash out how add this into CrocoDash and work on the parameter changes a bit. That discussion should be in a seperate PR so we can roll forward with this one independent of the CICE specific stuff.

PR #106, opened 2025-08-14
CESM Global to OBC Module

manishvenu:
- Using lessons from BGC write a module to grab a dataset, and create OBC data. This can almost take the place of regional_mom6 if we do this right.


manishvenu:
Ready for a first round of review. I'll run this on the MARBL BGC and try to regenerate the tracers we do have, redo the tests, create the test fixtures, write the documentation, & figure out how we plug this into CrocoDash case.

PR #108, opened 2025-08-15
Add Runoff + Finalize Hung's CICE to CrocoDash

manishvenu:
Changes:

1. Steps added to Case
2. Raw Data Product added for CESM Inputdata - TBD, once the runoff product is finalized

Co-authored-by: @silverfoxpt

### DIFF
--- a/CrocoDash/case.py
+++ b/CrocoDash/case.py
@@ -346,14 +346,7 @@ class Case:
         --------
         process_forcings : Executes the actual boundary, initial condition, and tide setup based on the configuration.
         """
-        if self.runoff_in_compset and (runoff_esmf_mesh_filepath is None):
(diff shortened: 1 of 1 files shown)


manishvenu:
@alperaltuntas, 

We talked about "./xmlchange ROF_DOMAIN_MESH=...
./xmlchange ROF_NX=..
./xmlchange ROF_NY=...

In the long term, this should be done automatically by visualCaseGen, when it adds new resolutions to modelgrid_aliases_nuopc.xml. Currently, it only sets atm, lnd and ocnice grids for a new resolution:"


 However, should this be done through VCG? We anyway need the user to grab the ESMF Mesh for the mapping, so we may as well set it as XML vars.

alperaltuntas (approved):

PR #114, opened 2025-09-22
Add River Nutrients Mapping + (Empty) Iron Forcing Generation for BGC

manishvenu:
Changes:

1. Offline mapping for a passed-in river nutrients file w/ Unit Conversion & Adding a timestep
2. Generates an empty file for iron forcing files for bgc runs

Background:

This uses the "configure" and "process" workflow.

### DIFF
--- a/CrocoDash/case.py
+++ b/CrocoDash/case.py
@@ -26,6 +26,8 @@ from visualCaseGen.custom_widget_types.case_creator import CaseCreator, ERROR, R
 from visualCaseGen.custom_widget_types.case_tools import xmlchange, append_user_nl
 from mom6_bathy import chl, mapping, grid
 import xesmf as xe
+import xarray as xr 
+import numpy as np
 
 class Case:
     """This class represents a regional MOM6 case within the CESM framework. It is similar to the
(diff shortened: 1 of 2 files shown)


CrocoDash/case.py
```diff
@@ -579,47 +594,65 @@ def process_forcings(
...
+                    np.zeros((depth, ny, nx), dtype=np.float32),
+                ),
+                "FESEDFLUXIN_oxic": (
+                    ["DEPTH", "ny", "nx"],
+                    np.zeros((depth, ny, nx), dtype=np.float32),
+                ),
+                "FESEDFLUXIN_reduce": (
+                    ["DEPTH", "ny", "nx"],
+                    np.zeros((depth, ny, nx), dtype=np.float32),
+                ),
```
mnlevy1981, line 626: I think I pointed out the attributes in the previous review, but these variables can be removed from the dataset (and it looks like the attributes are still being defined below)

mnlevy1981: Sorry, I missed the comment you added. I do think it's okay to remove these; when we add in the actual forcing we can talk about whether we want to bring them back or not. The proposed forcing in https://github.com/marbl-ecosys/MARBL/pull/464 separates the oxidized and reduced fluxes into two files rather than adding them together for a total flux, and the vent flux isn't split in the same way so it's a little misleading to see those terms in the file.

manishvenu:
Should be wrapped up! Will merge in a couple hrs if no issues

PR #165, opened 2025-12-08
Restructure Configure Forcings

manishvenu:
**Details:**
For shareable CrocoDash cases, it'll help to have modular configurations, which means they should become classes. In this PR,I implement modular configurations similar to the raw_data_access module with a registry, ForcingConfigRegistry, to store the configurations and manage them, and Configurator classes which hold all the configurations (Tides, Chl, etc..)

**Changes:**
A new file called Forcing_configurations that holds the registry (FCR), the Configuration classes, which inherit from the BaseConfigurator abstract class, and helper parameter classes (input and output).

How it works is that the function configure_forcings pass in all its keyword arguments and the compset to the FCR. The FCR (in the main function, find_active_configurators) analyzes the inputs and the compset and lines up the required configuration classes. (For example,  a BGC compset needs the BGC configuration class, and if you pass in the tidal arguments like tpxo_filepath and such, it will line up the tides configuration class).

It then runs these configurations with run_configurations.

This makes testing a lot easier and makes it a lot more modular. It is more overhead though.

**Prep work for shareable configuration**
I also included everything I think we'll need for shareable configuration, which includes serialization & configuration "inspection" which can read from user_nl & xmlquery.

To implement this, on top of just configuring the forcing, the configuration classes need to be "aware" of the changes they make, so input and output parameters are declared seperately in each class, in two class variables called input & output params. These are used in `__init__`, apply (which is the actual configuration), serialize, and inspect.

There may not be a reason for both serialization and inspection (we may want to get rid of inspection), but I thought I'd add it in for now.

Some assumptions:
1. We can access the caseroot though cvars
2. The eventual goal of shareable configurations does not mean you can roll back a case, you can only clone & modify

You can see the planning document for this PR and the shareable configuration here: https://docs.google.com/document/d/1PyZ6rZkhYxAx5A-0J-5gZid1cwNs5DZDhsUvV-8-36w/edit?usp=sharing

Minor Changes:
I refactored utils into logging!
Minor bug in testing w/ the Docker image.

Please see the docs link here for how it works! https://crocodile-cesm.github.io/CrocoDash/pr-165/forcing_configurations.html

### DIFF
--- a/tests/test_workflow.py
+++ b/tests/test_workflow.py
@@ -76,11 +76,11 @@ def run_full_workflow(
             bounds["ic"]["lon_min"],
             bounds["ic"]["lon_max"],
         )
-        ds.to_netcdf(case.inputdir / "glorys" / "extract_forcings"/"raw_data"/ "ic_unprocessed.nc")
-        ds.to_netcdf(case.inputdir / "glorys" / "extract_forcings"/"raw_data" / "east_unprocessed.20200101_20200201.nc")
-        ds.to_netcdf(case.inputdir / "glorys" / "extract_forcings"/"raw_data" / "west_unprocessed.20200101_20200201.nc")
-        ds.to_netcdf(case.inputdir / "glorys" / "extract_forcings"/"raw_data" / "north_unprocessed.20200101_20200201.nc")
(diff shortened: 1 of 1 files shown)


manishvenu:
Docs on this PR: https://crocodile-cesm.github.io/CrocoDash/pr-165/forcing_configurations.html

PR #173, opened 2026-01-08
Restructure Process Forcings

manishvenu:
Details:

The next step from restructuring the configure forcings I thought was useful is making the process_forcings functions into their own scripts. They are usually very simple functions, but some do have more complicated logic that calls xarray and such things. That shouldn't be in case. 

The solution is that we already have a workflow called extract_forcings, that takes serialized configure_forcings output for initial/boundary condition output.. So we just expand that to include all the other forcings like chl, tides, etc...

This is great because it solves the above problem, becomes much easier to test, and makes the logical flow of CrocoDash much simpler in my opionion. CrocoDash `case.py` instantiates your case and manages the workflow. `forcings_configurations` submodule configures your forcings. `extract_forcings` module processes your forcings. `raw_data_access` module gets all your raw data.

Changes:
1. Created scripts in `extract_forcings` for each `process_forcings` function
2. Redid `driver.py` to be more command-line favorable. Now you can do things like `python driver.py --all` or `python driver.py --runoff` or `python driver.py --all --skip conditions bgcic`
3. Added the VCG Update Here, and the newest RM6 commit that we use.
4. Modified `case.py` to accommodate these changes
6. Debug some issues in forcing_configuration
7. `extract_forcings` used to copy the processing code with the case specific files. That's not necessary (It was originally thought it could be useful to modify the code if you're processing things slightly differently, but it mostly just increases development time). So it's refactored to only copy the case setup (driver & config) and not the code. Now `extract_forcings` has the code at the first level, and then a subfolder called `case_setup` that gets copied into your case input directory. 
8. Removed `too_much_data`. Decided we don't need to get the data from `configure_forcings`. It's part of `process_forcings` now.
9. Refactored forcing_configurations into its own module to kinda go on-theme with our submodule layout. (also split base from actual configurations)

Notes:

1. This looks like a lot, and it is, but at its core, this is just a refactor, there is no new code development here other than changes to python driver.py
2. Docs that may be useful: https://crocodile-cesm.github.io/CrocoDash/pr-174/for_users/extract_forcings.html


CrocoDash/extract_forcings/case_setup/driver.py
```diff
@@ -0,0 +1,262 @@
...
+            date_format=config["basic"]["dates"]["format"],
+            start_date=config["basic"]["dates"]["start"],
+            end_date=config["basic"]["dates"]["end"],
+            hgrid_path=config["basic"]["paths"]["hgrid_path"],
+            step_days=int(config["basic"]["general"]["step"]),
+            output_dir=config["basic"]["paths"]["raw_dataset_path"],
+            boundary_number_conversion=config["basic"]["general"][
+                "boundary_number_conversion"
+            ],
+            run_initial_condition=run_initial_conditions,
```
alperaltuntas: Where does`run_initial_conditions` come from? I don't see it anywhere else.

CrocoDash/extract_forcings/case_setup/driver.py
```diff
@@ -0,0 +1,262 @@
...
+
+
+def run_from_cli(args):
+    cfg = utils.Config(CONFIG_PATH)
+
+    if args.test:
+        test_driver()
+        return
+
+    run_ic, run_bc = resolve_conditions(args)
```
alperaltuntas: I can't find `resolve_conditions` anywhere. How does this work? Were these changes run or tested at all?

CrocoDash/extract_forcings/case_setup/driver.py
```diff
@@ -0,0 +1,262 @@
...
+    # Conditions pipeline is special (comes from "basic")
+    if (args.all or args.conditions) and "conditions" not in (args.skip or []):
```
alperaltuntas: `args.conditions` and `args.skip` are referenced in multiple places, but they seem to be undefined. Judging by this and the other issues above, this PR doesn’t seem ready for review yet. Could you run/test/review this new code, fix issues, and let me know once it’s in a reviewable state?

alperaltuntas (changes requested):

CrocoDash/extract_forcings/case_setup/driver.py
manishvenu: Added tests for this.

CrocoDash/extract_forcings/utils.py
```diff
@@ -4,43 +4,33 @@
...
+        )
+        self.inputdir = Path(self.config["basic"]["paths"]["input_dataset_path"])
 
-    Parameters
-    ----------
-    config : dict
-        Configuration dictionary to save.
-    config_path : str, optional
-        Path to the JSON config file. Default is "config.json".
+    def keys():
```
alperaltuntas: add self arg.

manishvenu:
in-person approved! This'll be merged

PR #181, opened 2026-03-12
Docs Iteration

manishvenu:
1) Rework docs into tutorial like format with steps
2) Write a few more tests
3) Minor Bug Fixes


alperaltuntas:
@manishvenu Looks like this PR got outdated again. Could you update it and fix the CI failures?

manishvenu:
> @manishvenu Looks like this PR got outdated again. Could you update it and fix the CI failures?

@alperaltuntas, The VCG commit was just out of date. It should be good to go now.

tests/fixtures/grid_gen.py
```diff
@@ -22,6 +22,7 @@ def get_rect_grid_and_empty_topo(get_rect_grid):
     topo = Topo(
         grid=get_rect_grid,
         min_depth=9.5,
+        version_control_dir=tmp_path / "TopoLibrary",
```
alperaltuntas: tmp_path is undefined.

tests/fixtures/grid_gen.py
```diff
@@ -31,6 +32,7 @@ def get_rect_grid_and_topo(get_rect_grid):
...
```
alperaltuntas: tmp_path is undefined.

.github/workflows/deploy_sphinx_docs.yml
```diff
@@ -16,6 +16,56 @@ jobs:
       - uses: actions/checkout@v4
         with:
             submodules: recursive
+      - name: Check API docs are up to date
```
alperaltuntas, line 21: This is a good addition, but please double check if it's exercised end to end. Also, did you confirm these new steps actually skip the env rebuild when environment.yml is unchanged? Worth comparing this checkout against how testing.yml sets up its checkout for the same container image: if the two differ, and if a $WORKDIR/$GITHUB_WORKSPACE path assumption doesn't carry over, this will fail quietly rather than error.

alperaltuntas (changes requested):

alperaltuntas (approved):

PR #182, opened 2026-03-17
Diag Manager UI

### LINKED ISSUE #188: Diag Table Modifier
Using Anythony's package. Will debate if we want this. https://github.com/CROCODILE-CESM/CrocoDash/pull/182

manishvenu:
Details:
I [forked Anthony's diag manager](https://github.com/CROCODILE-CESM/mom6_diagnostics_gui) and "slashed and burned" it a bit so I could handle the complexity. I then added in as an available option here as a function case.manage_diags. The dependecies are just fine because it is just an ipywidget, which is great.

Closes #188


manishvenu:
@AidanJanney Is going to a build a mom6-tools equivalent that is better for our use case.

PR #207, opened 2026-04-02
CrocoGallery Testing w/ Cirrus Bugs Fixing

manishvenu:
A new workflow to test the notebooks


manishvenu:
Hey @alperaltuntas, Looks like the runner is taking forever to pick up the job, so I'm just marking this as ready because I think it is!

manishvenu:
See a sample run of this new github action in the commit before the most recent.

alperaltuntas (approved):

PR #218, opened 2026-05-20
Refactor OBC forcing pipeline

### LINKED ISSUE #185: Optimize extract_forcings with Dask
The OBC workflow is chunked but not parallelized!

manishvenu:
## What changed

This PR started as a Dask parallelism experiment for the OBC GET (download) step, and is being merged as a **sequential-only refactor** after the parallelism approach hit fundamental design blockers.

### Why parallelism doesn't work at the orchestration level

**Dask distributed (process workers)** hit two independent blockers:
1. `ESMCI::VM::getCurrent()` returns rc=545 in any child process on PBS — REGRID is forced into the main process regardless, so only GET could be parallelized.
2. `copernicusmarine` opens S3-backed zarr stores and calls `dask.compute()` internally during `to_netcdf()`. The distributed scheduler tries to serialize botocore's `S3Client` across process boundaries, which fails.

**`ThreadPoolExecutor`** looks like the obvious fix — but GET functions in the registry are not uniformly I/O-bound. Some read from local datasets and do CPU work (subsetting, format conversion). Threads don't parallelize CPU-bound work due to the GIL, so there's no single threading model that works cleanly across all products.

**The answer:** each product's GET function should own its own parallelism internally, since only it knows whether its bottleneck is network, disk, or CPU. The orchestration layer has no basis for making that decision.

**Changes:**

**Decoupled GET and REGRID chunking** — GET and REGRID now have independent step sizes in the config (`get_step` and `regrid_step`). GET defaults to the full date range in one request (driven by data-provider constraints); REGRID defaults to 30-day slices (driven by memory/xESMF performance). After GET, raw files are opened lazily with `open_mfdataset` and sliced by `regrid_step` independently of how GET chunked the download.

**Filename-based coverage validation** — after GET, raw file coverage is validated by parsing dates from filenames (`east_unprocessed.2020-01-01_2020-01-31.nc`) without opening any files. Checks for correct start/end, gaps, and overlaps.

**Naming conventions out of config** — `file_regex` is removed from `config.json`. The raw and regridded filename formats are hardcoded in `obc.py` where they are actually used, not in the config.

**Flexible date parsing** — config dates are stored as ISO strings (`2020-01-01`). The `"format"` key is gone. All date parsing uses `pd.to_datetime()` / `datetime.fromisoformat()` — any standard date string works.

**Cleaner config structure** — the `config.json` written by `configure_forcings` is built as a single dict literal (no two-phase init/overwrite pattern). The removed keys are: `"format"`, `"file_regex"`.

**Idempotent pipeline** — GET, REGRID, and MERGE are each idempotent. Existing output files are detected and skipped at each phase. Safe to re-run after a partial failure.

**Regridder weight reuse** — regridder weights are computed once on the first REGRID chunk per boundary and reused for all subsequent chunks, avoiding redundant xESMF setup.

**Cleaner driver.py** — `run_workflow()` is the shared core used by both the CLI and `case.process_forcings()`. All cluster/Dask CLI flags removed.

The user-facing API (`case.process_forcings()`, `process_obc_conditions(config_path)`) is unchanged.

Closes #185


alperaltuntas (commented):
This PR is failing the CI as well. Also, extract_forcings.md still describes a parallelism API.

CrocoDash/extract_forcings/obc.py
```diff
@@ -0,0 +1,455 @@
...
+                raise RuntimeError(
+                    f"Regridded file {dated_output} exists but is not valid NetCDF. "
+                    "Delete it and re-run."
+                )
+            logger.info(f"Regridded file {dated_output.name} already exists. Skipping.")
+            regridded_files.append(dated_output)
+            continue
+
+        tmp_file = output_folder / f"_tmp_{boundary}_{start_str}_{end_str}.nc"
+        ds_full.sel(time=slice(chunk_start, chunk_end)).to_netcdf(tmp_file)
```
alperaltuntas: chunk_start/chunk_end come from _make_date_pairs, which works purely in whole calendar days. What's the actual timestamp convention on the raw product's time coordinate? Worth tracing through what this `.sel(time=slice(...))` actually keeps at each chunk boundary.

CrocoDash/extract_forcings/obc.py
```diff
@@ -0,0 +1,455 @@
...
+            function_name=function_name,
+            variables=variables,
+            extra_args=extra_args,
+        )
+
+    regridded_files_by_boundary = {}
+    for boundary in boundaries:
+        seg_id = bnc[boundary]
+        raw_files = _validate_coverage(
+            sorted(raw_path.glob(f"{boundary}_unprocessed.*.nc")),
```
alperaltuntas: This glob picks up anything matching the boundary prefix. If someone reruns this against a raw_dataset_path that already has output from the previous run with a different date range, what does _validate_coverage see?

CrocoDash/extract_forcings/obc.py
```diff
@@ -0,0 +1,455 @@
...
+                )
+            logger.info(f"OBC file {output_file.name} already exists. Skipping.")
+            continue
+
+        ProductRegistry.load()
+        data_access_fn = ProductRegistry.get_access_function(
+            product_name, function_name
+        )
+        hgrid = xr.open_dataset(hgrid_path)
+        latlon = Grid.get_bounding_boxes(hgrid)[boundary]
```
alperaltuntas: Are these lines repeated for each chunk? If so, isn't it redundant?

CrocoDash/extract_forcings/initial_condition.py
```diff
@@ -0,0 +1,237 @@
...
+    hgrid = xr.open_dataset(hgrid_path)
+    boundary_info = Grid.get_bounding_boxes(hgrid)
+    latlon_info = boundary_info["ic"]
+    output_file = "ic_unprocessed.nc"
+    end_ic_date = start_date + timedelta(days=1)
+    end_ic_date_str = end_ic_date.strftime("%Y-%m-%d")
+    start_date_str = start_date.strftime("%Y-%m-%d")
+
+    # Build requested variables
+    phys_vars = [
```
alperaltuntas: This is repeated in obc.py. DRY please.

CrocoDash/extract_forcings/initial_condition.py
```diff
@@ -0,0 +1,237 @@
...
+            variables=phys_vars + extra_tracers,
+            extra_args=extra_args,
+        )
+
+    # Set up required information
+    expt = rm6.experiment.create_empty()
+    expt.hgrid = hgrid
+    expt.mom_input_dir = Path(output_data_dir)
+    expt.date_range = [start_date, None]
+    vgrid_from_file = xr.open_dataset(vgrid_path)
```
alperaltuntas, line 82: The old regrid_dataset_piecewise.py had an explicit FileNotFoundError with a pointed message when vgrid_path was missing for an IC run. What does a user actually see now if they call this without a valid vgrid_path?

alperaltuntas (changes requested):

CrocoDash/extract_forcings/obc.py
manishvenu: This is interesting because I just ran into this. It's midnight of that day so the end date is non inclusive. I will put a fix now to make sure end date is inclusive

CrocoDash/extract_forcings/obc.py
manishvenu: It fails with overlapping dates.

CrocoDash/extract_forcings/obc.py
manishvenu: Yes, missed that, moved it out of the loop

CrocoDash/extract_forcings/initial_condition.py
manishvenu: This would fail with a missing file error when you call xr.open_dataset, but this isn't something that's provided by the user, it's a case variable that's written out by forcing configurations, so hopefully it is never manually passed in by the user. Do you think we still need the pointed message?

CrocoDash/extract_forcings/obc.py
manishvenu: So I decided to have the raw_data_access module enforce that forcing products must be inclusive on the ending date (which is in units of days)

alperaltuntas: @manishvenu This still comes back to the timestamp convention I asked about earlier. `_make_date_pairs` gives you midnight-to-midnight day pairs, and `.sel(time=slice(...))` is inclusive on both ends, so if the product stamps its daily fields at noon, the last day of each chunk falls past chunk_end and the next chunk starts the day after, which disappears. And since `_validate_coverage` reads dates off the filenames rather than the data, it'd still report full coverage. Can we trace what the raw product actually puts on the time axis, and make sure a day can't slip between two chunks?

CrocoDash/extract_forcings/initial_condition.py
alperaltuntas: We used to raise a clear FileNotFoundError here when vgrid_path was missing on an IC run. Now it's a plain xr.open_dataset(vgrid_path), so a missing or None path just gives the user a raw xarray traceback. Can we bring the explicit check and message back?

CrocoDash/extract_forcings/obc.py
alperaltuntas: The endpoint/gap checks in _validate_coverage do catch a mismatched range now, which is nice. But the glob still grabs everything matching the boundary prefix, so rerunning into a raw_dataset_path that already has a different date range now errors out, which is a little at odds with the "safe to re-run" note in the docstring. Should we scope the glob to the requested range, or at least call out that the raw dir is single-range?

CrocoDash/extract_forcings/obc.py
```diff
@@ -0,0 +1,414 @@
...
+                )
+            logger.info(f"Regridded file {dated_output.name} already exists. Skipping.")
+            regridded_files.append(dated_output)
+            continue
+
+        tmp_file = output_folder / f"_tmp_{boundary}_{start_str}_{end_str}.nc"
+        ds_full.sel(time=slice(chunk_start, chunk_end)).to_netcdf(tmp_file)
+
+        try:
+            hgrid = xr.open_dataset(hgrid_path)
```
alperaltuntas: Reusing the regridder weights across chunks looks good. xr.open_dataset(hgrid_path) and the rm6.segment(...) are still rebuilt every chunk though. Could the hgrid open move above the loop, like you did in the GET path?

alperaltuntas:
@manishvenu Please address the unresolved comments above.

CrocoDash/extract_forcings/obc.py
manishvenu: I moved the hgrid open, segment needs to be init every time because it takes a date, but that will change in future rm6 dev.

CrocoDash/extract_forcings/initial_condition.py
manishvenu: Done!

CrocoDash/extract_forcings/initial_condition.py
manishvenu: Should be resolved.

alperaltuntas:
@manishvenu Please address the unresolved comments above.

CrocoDash/extract_forcings/obc.py
manishvenu: I'm worried that opening every file to read the time axis is going to be really costly, so would it be possible to do a more elaborate validate converge later? In the meantime I've made sure the chunk_end is to the last second of the day.

CrocoDash/extract_forcings/obc.py
manishvenu: Yes, I added a simple less than greater than func after the glob that then goes into the validate coverage

manishvenu:
> @manishvenu Please address the unresolved comments above.

Hey @alperaltuntas, I think I actually resolved them now! Completely missed the two above comments.

alperaltuntas (approved):

PR #220, opened 2026-05-26
Flexible Calendar Assignment based on ForcingProduct

### LINKED ISSUE #35: CESM Case Fails with Leap Years
When running with a time frame that includes a leap year, the CESM run fails with the following diagnostic (this example run started 2000-01-01): 
`FATAL from PE     0: get_cal_time: Error in function get_cal_time: Invalid date. Date=2000-02-29 00:00:00 Note that the time_manager is using the NOLEAP calendar while the calendar type passed to function get_cal_time is julian`

Manish helped me find a workaround by manually changing this with: `./xmlchange CALENDAR='GREGORIAN'` from the original `CALENDAR='NO_LEAP'`

This could be added as another step in case.py, but there might also be good reasons to keep the 'NO_LEAP' setting generally. We could maybe change it to Gregorian conditionally if the time domain contains a leap year?

### LINKED ISSUE #150: [feature] Default MOM6 calendar to Gregorian
When comparing MOM6 output with real obs or when performing DA, MOM6 needs to be run with Gregorian calendar, else the observation date and the model date don't match.

From the CLI this can be done in CESM with:
```
./xmlchange CALENDAR=GREGORIAN
```

so my suggestion would be to set this as the default in CrocoDash somewhere, given that within the Crocodile project we do DA and model-obs comparison. I don't know why the default is different and if there's a good reason to keep that, in which case I would suggest to provide a tool/flag/argument to pass on the Gregorian option

### LINKED ISSUE #246: GLORYS dataset is Gregorian
When doing the setup we just need to do:

./xmlchange CALENDAR=GREGORIAN

before the build.

manishvenu:
In configure_forcings, check if any Feb 29 falls within the date range and apply the xmlchange if so.

Closes #35, Closes #150, Closes #246

I'll open an issue to make sure we come back to this at some point.


CrocoDash/case.py
```diff
@@ -459,6 +459,12 @@ def configure_forcings(
         self.fcr.run_configurators(self.extract_forcings_path / "config.json")
 
         self._update_forcing_variables()
+
+        start, end = pd.to_datetime(date_range[0]), pd.to_datetime(date_range[1])
+        dates = pd.date_range(start, end, freq="D")
+        if any((d.month == 2 and d.day == 29) for d in dates):
+            xmlchange("CALENDAR", "gregorian", is_non_local=self.cc._is_non_local())
```
alperaltuntas: Three are `calendar` arguments/properties scattered across the software stack, including in 

```
./extract_forcings/bgc.py
./rm6/regional_mom6/regional_mom6.py
./raw_data_access/datasets/mom6_output.py
./visualCaseGen/external/mom6_forge/mom6_forge/chl.py:
etc.
```

How do we ensure consistency across all of these different calendar specifications in the software stack, including with this new addition?

enrico-mi:
> In configure_forcings, check if any Feb 29 falls within the date range and apply the xmlchange if so.
> 

Hey @manishvenu , I saw this while checking the status of #150 .

It'd be great if Gregorian calendar is always the default, not only when the simulated time includes a Feb 29 date: CESM stores the 'time' coordinate of history files differently if GREGORIAN or NOLEAP is used (i.e. with/without `ds['time'].attrs='gregorian'`), so the user would see an inconsistent behaviour depending on the time range they simulate. Pipelines that depend on the calendar also break because of this (e.g. model2obs that time-matches MOM6 values and external obs).

CrocoDash/case.py
manishvenu: Expose all the calendar arguments to the user. I've done most of them here. 

https://github.com/NCAR/mom6_forge/pull/112

manishvenu:
> > In configure_forcings, check if any Feb 29 falls within the date range and apply the xmlchange if so.
> 
> Hey @manishvenu , I saw this while checking the status of #150 .
> 
> It'd be great if Gregorian calendar is always the default, not only when the simulated time includes a Feb 29 date: CESM stores the 'time' coordinate of history files differently if GREGORIAN or NOLEAP is used (i.e. with/without `ds['time'].attrs='gregorian'`), so the user would see an inconsistent behaviour depending on the time range they simulate. Pipelines that depend on the calendar also break because of this (e.g. model2obs that time-matches MOM6 values and external obs).

Should be handled now!

enrico-mi:
> > > In configure_forcings, check if any Feb 29 falls within the date range and apply the xmlchange if so.
> > 
> > 
> > Hey @manishvenu , I saw this while checking the status of #150 .
> > It'd be great if Gregorian calendar is always the default, not only when the simulated time includes a Feb 29 date: CESM stores the 'time' coordinate of history files differently if GREGORIAN or NOLEAP is used (i.e. with/without `ds['time'].attrs='gregorian'`), so the user would see an inconsistent behaviour depending on the time range they simulate. Pipelines that depend on the calendar also break because of this (e.g. model2obs that time-matches MOM6 values and external obs).
> 
> Should be handled now!

So if I understand the codebase correctly, the default is noleap but whenever CrocoDash generates a case it runs xmlchange and makes it Gregorian, right? that works.

Do you want me to test it?

manishvenu:
> > > > In configure_forcings, check if any Feb 29 falls within the date range and apply the xmlchange if so.
> > > 
> > > 
> > > Hey @manishvenu , I saw this while checking the status of #150 .
> > > It'd be great if Gregorian calendar is always the default, not only when the simulated time includes a Feb 29 date: CESM stores the 'time' coordinate of history files differently if GREGORIAN or NOLEAP is used (i.e. with/without `ds['time'].attrs='gregorian'`), so the user would see an inconsistent behaviour depending on the time range they simulate. Pipelines that depend on the calendar also break because of this (e.g. model2obs that time-matches MOM6 values and external obs).
> > 
> > 
> > Should be handled now!
> 
> So if I understand the codebase correctly, the default is noleap but whenever CrocoDash generates a case it runs xmlchange and makes it Gregorian, right? that works.
> 
> Do you want me to test it?

Hey Enrico, The idea is that your calendar is set to whatever calendar the forcing product is in. So for glorys it's GREGORIAN, but CESM output it's no leap. But I'm starting to think I didn't understand what you were requesting....

If you have the time to test this that would be awesome, but no sweat either way.

enrico-mi:
> > > > > In configure_forcings, check if any Feb 29 falls within the date range and apply the xmlchange if so.
> > > > 
> > > > 
> > > > Hey @manishvenu , I saw this while checking the status of #150 .
> > > > It'd be great if Gregorian calendar is always the default, not only when the simulated time includes a Feb 29 date: CESM stores the 'time' coordinate of history files differently if GREGORIAN or NOLEAP is used (i.e. with/without `ds['time'].attrs='gregorian'`), so the user would see an inconsistent behaviour depending on the time range they simulate. Pipelines that depend on the calendar also break because of this (e.g. model2obs that time-matches MOM6 values and external obs).
> > > 
> > > 
> > > Should be handled now!
> > 
> > 
> > So if I understand the codebase correctly, the default is noleap but whenever CrocoDash generates a case it runs xmlchange and makes it Gregorian, right? that works.
> > Do you want me to test it?
> 
> Hey Enrico, The idea is that your calendar is set to whatever calendar the forcing product is in. So for glorys it's GREGORIAN, but CESM output it's no leap. But I'm starting to think I didn't understand what you were requesting....
> 
> If you have the time to test this that would be awesome, but no sweat either way.

No problems, I'll test it. Thanks!

CrocoDash/case.py
```diff
@@ -453,12 +453,22 @@ def configure_forcings(
             "boundaries": boundaries,
         }
 
+        forcing_product = ProductRegistry.get_product(self.forcing_product_name.lower())
+
         self.session_id = cvars["MB_ATTEMPT_ID"].value
         self.grid_name = self.ocn_grid.name
+        self.forcing_product = forcing_product
```
alperaltuntas, line 453: Why do we add a `forcing_product` member?

CrocoDash/raw_data_access/datasets/glorys.py
```diff
@@ -38,6 +38,8 @@ class GLORYS(ForcingProduct):
     eta_var_name = "zos"
     depth_coord = "depth"
     tracer_var_names = {"temp": "thetao", "salt": "so"}
+    cf_calendar = "julian"
```
alperaltuntas: Two separate variables for the same calendar is error prone. And I think there is already a bug here related to that: cf_calendar = "julian" is inconsistent with cesm_calendar= "GREGORIAN".

Please consider a safer approach. Here is a one possible solution:

```
@dataclass(frozen=True)
class Calendar:
    cf: str      # CF name for dataset time attrs
    cesm: str    # CIME CALENDAR xml value

GREGORIAN = Calendar(cf="standard", cesm="GREGORIAN")
NOLEAP = Calendar(cf="noleap", cesm="NO_LEAP")
```

In ForcingProduct.__init_subclass__, before super().__init_subclass__(...):
```
calendar = getattr(cls, "calendar", None)
if calendar is not None:
    cls.cf_calendar = calendar.cf
    cls.cesm_calendar = calendar.cesm
```

Then each product declares one line `calendar = GREGORIAN` and the two representations can't disagree.

alperaltuntas (changes requested):

CrocoDash/raw_data_access/datasets/glorys.py
manishvenu: I had though the glorys data was in julian for some reason, but it is gregorian. I'll fix that.

manishvenu: I like this solution! I'll do it this way.

CrocoDash/case.py
manishvenu: So it can be passed into the forcing_configurations.

manishvenu: So it can be used in "case_forcing_product=None,"

alperaltuntas (approved):

PR #222, opened 2026-05-28
SRTM & CESM LENS & Generic CESM MOM/POP Raw Data!

### LINKED ISSUE #52: Add Data Access Product: SRTM Bathymetry

### LINKED ISSUE #76: Better guidance on CrocoDash environment requirement for running get_glorys_data.sh
When the user forgets to activate CrocoDash conda environment, the get_glorys_data script prints out:

```
get_glorys_data.sh: line 3: copernicusmarine: command not found
get_glorys_data.sh: line 6: copernicusmarine: command not found
get_glorys_data.sh: line 9: copernicusmarine: command not found
get_glorys_data.sh: line 12: copernicusmarine: command not found
get_glorys_data.sh: line 15: copernicusmarine: command not found
```

due to the fact that copernicusmarine is available via the CrocoDash conda environment. We should let users know that they need to activate the environment. For instance, we can add the following if statement to the beginning of the auto-generated get_glorys_data.sh files:

```
if [[ "$CONDA_DEFAULT_ENV" != "CrocoDash" ]]; then
    echo "Error: Please activate the 'CrocoDash' conda environment before running this script. To do so, run:"
    echo "conda activate CrocoDash"
    exit 1
fi
```

### LINKED ISSUE #190: More forcing products! CESM LE for the most part

manishvenu:
Changes:
1. Add SRTM bathymetry support.
2. Rename `mom6_output.py` → `cesm_ocean_output.py`.
3. Add CESM2-LENS2 tseries support (`member=` support alongside CESM-HR FOSI).
4. Split the single `CESM_OCEAN_OUTPUT` product into two products with correct, product-specific metadata:
   - `CESM_POP_OUTPUT` — POP-grid CESM tseries output (FOSI, LENS2). Single method `get_cesm_single_variable_data` (merged from the old `get_pop_data`/`get_cesm2_lens_data`).
   - `CESM_MOM_OUTPUT` — native MOM6 output, with MOM6-native variable/coordinate metadata (`thetao`/`so`/`uo`/`vo`/`zos` on `xh`/`yh`/`xq`/`yq`/`z_l`). Two methods:
     - `get_mom6_output_data` — generic reader for native MOM6 output (full history/diagnostic files, or diag_table-extracted cross-section slices) pointed at any directory, for e.g. nesting a child domain inside a parent/outer-nest MOM6 run. Supports both full-domain and pre-sliced input through the same bbox-mask code path.
     - `get_mom6_single_variable_data` — same generic MOM6 support, but for MOM6 output organized in the CESM single-variable-per-file (tseries) convention.
5. Fixed a performance bug in `parse_dataset`: it previously did a full recursive `rglob` + per-file `stat()` over entire CESM archive directories, taking ~11 minutes against real FOSI/LENS2 archives. Now scopes the scan to each variable's own subdirectory when present and uses `os.walk` for the flat-archive fallback — cut real-archive runtime to ~12 seconds.
6. Fixed a thread-safety issue: `xr.open_mfdataset(..., parallel=True)` is unsafe with the netCDF4 backend for concurrent multi-file reads (could silently drop/corrupt variables or crash) — removed `parallel=True` from the tseries merge step.

Supersedes #249 (a narrower single-purpose nesting reader) — that approach was folded into the generic `CESM_MOM_OUTPUT.get_mom6_output_data` here instead.

Known limitation: `CESM_MOM_OUTPUT` has no universal default `dataset_path` (it's meant to point at any user-supplied directory), so using it through the fully-automated `case.configure_forcings()` pipeline currently requires setting `dataset_path` as a class attribute beforehand.
Closes #190 Closes #52 Closes #76


CrocoDash/raw_data_access/datasets/cesm_ocean_output.py
```diff
@@ -0,0 +1,715 @@
...
+        dates: list,
+        lat_min,
+        lat_max,
+        lon_min,
+        lon_max,
+        name=None,
+        output_folder=Path(""),
+        output_filename=None,
+        variables=["zos", "thetao", "so", "uo", "vo"],
+        dataset_path="/glade/u/home/manishrv/scratch/archive/gibraltar_parent/ocn/hist/z_files",
```
alperaltuntas: Do we really need to provide a default path here? If so, can we have a more permanent directory as the default `dataset_path` rather than what looks like a scratch directory?

CrocoDash/raw_data_access/datasets/cesm_ocean_output.py
```diff
@@ -0,0 +1,715 @@
...
+    # fresh per variable (not cached/reused across the loop) - variables can
+    # sit on different horizontal grid points with different lat/lon dims.
+    output_file_paths = []
+    for var_name, file_paths in variable_info.items():
+        if dates is None:
+            dates = ("NotSpecifiedDate", "NotSpecifiedDate")
+        output_file = output_path / (
+            f"{var_name}_subset_{lat_min}_{lat_max}_{lon_min}_{lon_max}_{dates[0]}_{dates[1]}.nc"
+        )
+        output_file_paths.append(output_file)
```
alperaltuntas: This appends the path before the checks below that skip a variable with no matching files, so a requested variable with zero files contributes a path that never gets written. When output_filename is set, read_single_variable_tseries_data feeds this list into xr.open_mfdataset for the merge, which then fails with FileNotFoundError on the missing file (same when preview is True).

alperaltuntas (changes requested):

CrocoDash/raw_data_access/datasets/cesm_ocean_output.py
manishvenu: Replaced to "please_provide_a_path"

CrocoDash/raw_data_access/datasets/cesm_ocean_output.py
manishvenu: Moved to the bottom.

CrocoDash/raw_data_access/datasets/cesm_ocean_output.py
alperaltuntas: Not sure if I like this solution. Can we set it to None? validate_dataset_path already handles None, so the only change needed is to give it a message that tells the user how to set the path rather than reporting that a fake path does not exist.

CrocoDash/raw_data_access/datasets/cesm_ocean_output.py
alperaltuntas: Moving the append to the end of the loop fixed the no-files case, but it introduced a regression here. This branch now continues before the append, so a subset that already exists on disk is dropped from the returned paths. On a re-run or resume, which is exactly what this skip is meant to support, those variables disappear from the list and the merge in read_single_variable_tseries_data gets an incomplete set or an empty list. Previously they were included. I think this branch should append before continuing, since the file does exist:

```
if output_file.exists():
    print(f"Subset already exists for {var_name}, skipping")
    output_file_paths.append(output_file)
    continue
```

Also worth noting the merge at read_single_variable_tseries_data still opens paths without checking existence, and with preview=True nothing is written yet the path is still appended, so preview plus output_filename would try to merge files that were never created. Filtering paths to existing files right before the merge would cover both.

CrocoDash/raw_data_access/datasets/cesm_ocean_output.py
```diff
@@ -0,0 +1,716 @@
...
+                file_path = Path(dirpath) / fname
+                matched = [
+                    v
+                    for v in candidate_vars
+                    if (space_character + v + space_character) in str(file_path)
+                ]
+                if not matched:
+                    continue
+                s = str(file_path.resolve())
+                dt1, dt2 = get_date_range_from_filename(s, regex)
```
alperaltuntas: get_date_range_from_filename returns None when the filename matches a variable token but not the date regex, and this line would then raise a TypeError trying to unpack None. Worth a guard like: `rng = get_date_range_from_filename(s, regex); if rng is None: continue.`

alperaltuntas:
@manishvenu Please see above my responses to the previous comments, as well as a new comment I just added.

CrocoDash/raw_data_access/datasets/cesm_ocean_output.py
manishvenu: Added both suggestions

CrocoDash/raw_data_access/datasets/cesm_ocean_output.py
manishvenu: Added this suggestion.

CrocoDash/raw_data_access/datasets/cesm_ocean_output.py
manishvenu: Added this suggestion

CrocoDash/raw_data_access/datasets/cesm_ocean_output.py
```diff
@@ -0,0 +1,730 @@
...
+                if not matched:
+                    continue
+                s = str(file_path.resolve())
+                rng = get_date_range_from_filename(s, regex)
+                if rng is None:  # no date range found in filename regex, skip this file
+                    continue
+                dt1, dt2 = rng
+                if (dt1 >= start_date and dt1 <= end_date) or (
+                    dt2 >= start_date and dt2 <= end_date
+                ):
```
alperaltuntas: This only keeps a file if one of its *endpoints* lands inside the window, so a file that spans the window (`dt1 < start_date and dt2 > end_date`) gets dropped. That's the normal case for the archives we're adding here: LENS2 is decadal chunks (`SALT.185001-185912.nc`), the default FOSI path is annual (`SALT.195801-195812.nc`). Ask for 1852-1855 and you get zero files — the variable goes silently missing from the returned paths, or you hit `OSError: no files to open` once `output_filename` is set. Should just be the usual overlap test:

```python
if dt1 <= end_date and dt2 >= start_date:
```

Carried over from `mom6_output.py` so it's not a regression, but LENS2 support makes it the common path rather than an edge case.

CrocoDash/raw_data_access/datasets/cesm_ocean_output.py
```diff
@@ -0,0 +1,730 @@
...
+        # Subset the dataset based on the provided geographical bounds
+        if not preview:
+            subset_ds = ds.where(mask, drop=True)
+
+            # Save the subsetted dataset to the output path
+
+            subset_ds.load().to_netcdf(output_file)
+
+            print(f"Subsetted dataset for variable '{var_name}' saved to {output_file}")
+        output_file_paths.append(output_file)
```
alperaltuntas: This append is outside the `if not preview` block, so preview mode returns paths to files that were never written. Anything using the return value rather than the datasets breaks on it. Can we move it inside the branch?

CrocoDash/raw_data_access/datasets/cesm_ocean_output.py
```diff
@@ -0,0 +1,730 @@
...
+    # Merge the file into the specified output file.
+    if output_filename is not None:
+        print(
+            f"Merging the files since output file is specified, into {Path(output_folder)/output_filename}"
+        )
+        # parallel=True is unsafe here: the netCDF4 backend isn't thread-safe
+        # for concurrent metadata reads across multiple files, and can
+        # silently drop/corrupt variables (or crash) under dask's threaded
+        # scheduler instead of raising a clear error.
+        existing_files = [p for p in paths if Path(p).is_file()]
```
alperaltuntas, line 446: The filter covers missing files but not an empty list — `xr.open_mfdataset([])` raises `OSError: no files to open`. That happens whenever no variable matched (see above), or when `preview` is True with `output_filename` set. Can we skip the merge when `existing_files` is empty?

CrocoDash/raw_data_access/datasets/cesm_ocean_output.py
```diff
@@ -0,0 +1,730 @@
...
+import cftime
+import dask.base
+import pandas as pd
+
+from CrocoDash.raw_data_access.base import *
+from CrocoDash.raw_data_access.datasets.utils import make_dates_end_inclusive
+
+
+class CESM_POP_OUTPUT(ForcingProduct):
+    product_name = "cesm_pop_output"
```
alperaltuntas, line 30: This rename breaks any existing `config.json` or shareable bundle naming `mom6_output`/`get_mom6_data` — the lookup fails with a bare `KeyError: 'mom6_output'`, no hint about the new name. Can we keep an alias, or make the registry error list the known products?

tests/raw_data_access/test_dates_inclusive.py
```diff
@@ -30,6 +30,11 @@
     ("glorys", "get_glorys_data_from_rda"),
     ("glofas", "get_global_data_with_python"),
     ("mom6_output", "get_mom6_data"),
```
alperaltuntas: Dead entry after the rename — this can't match a registered name anymore. Worth dropping.

alperaltuntas (changes requested):
Thanks, my three earlier comments are all addressed. A few new things from reading it end to end — the first three are blocking, the rest are notes.

One that isn't tied to a line: `products.csv` and `functions.csv` weren't regenerated, so the Datasets page still lists `mom6_output`/`get_mom6_data` and is missing `srtm`, `cesm_mom_output`, and the new access methods. `edit_docs.md` asks for these to be regenerated and committed: `cd docs/source && python raw_data_access/generate_info.py`

CrocoDash/raw_data_access/datasets/cesm_ocean_output.py
manishvenu: OK fixed!

CrocoDash/raw_data_access/datasets/cesm_ocean_output.py
manishvenu: Yes

CrocoDash/raw_data_access/datasets/cesm_ocean_output.py
manishvenu: Yes

CrocoDash/raw_data_access/datasets/cesm_ocean_output.py
manishvenu: Added a list of products to the get_product function.

tests/raw_data_access/test_dates_inclusive.py
manishvenu: Yup

alperaltuntas (approved):

PR #225, opened 2026-05-28
Write function default args to config.json for user overrides

manishvenu:
Details: opens the selected access function's non-required parameters, stamps their defaults into `config.json` under `forcing.function_args`, and passes them through to OBC and IC at runtime. Users can also pass `function_overrides` directly to `configure_forcings()` to override any of these defaults (e.g. `member` for CESM2-LENS2), instead of hand-editing `config.json` between `configure_forcings()` and `process_forcings()`.


CrocoDash/extract_forcings/initial_condition.py
```diff
@@ -25,6 +25,7 @@ def process_initial_condition(
     output_data_dir: str | Path,
     bathymetry_path: str | Path,
     preview: bool = False,
+    function_args: dict = None,
```
alperaltuntas, line 30: Where is this argument used?

CrocoDash/extract_forcings/case_setup/driver.py
```diff
@@ -308,6 +308,7 @@ def run_workflow(
             output_data_dir=cfg["basic"]["paths"]["output_path"],
             bathymetry_path=cfg["basic"]["paths"]["bathymetry_path"],
             preview=preview,
+            function_args=cfg["basic"]["forcing"].get("function_args", {}),
```
alperaltuntas, line 299: This passes function_args to the IC processor, but the `process_obc` call in the bc branch above doesn't, and `process_obc_conditions` never reads `forcing.function_args` from config. So boundary conditions never see user overrides even though the PR says they should.

alperaltuntas (changes requested):
@manishvenu In addition to the below line comments, please fix the CI failure.

CrocoDash/extract_forcings/initial_condition.py
manishvenu: This is fixed now

CrocoDash/extract_forcings/case_setup/driver.py
manishvenu: So now they should. It might have gotten lost in a merge.

alperaltuntas (approved):

PR #226, opened 2026-05-28
Use bathymetry tmask for OBC download bounding boxes

### LINKED ISSUE #164: Optimization request: configure_forcings file download
It looks like the configure_forcings step downloads forcing files for all boundary cells, including land points. We could probably speed things up by downloading data only for the wet (ocean) parts of the domain edges.

manishvenu:
**Descrip:**

Instead of using the full supergrid edge extent for each boundary download, compute per-boundary lat/lon bboxes from the bathymetry tmask — only requesting data over ocean cells. Closes #164.

**Changes:**

1. Per-boundary bboxes derived from the bathymetry tmask.

2. Falls back to full supergrid bounds if `bathymetry_path` is absent from config.


CrocoDash/extract_forcings/obc.py
```diff
@@ -20,12 +20,15 @@
...
 
+import dask
 import pandas as pd
 import regional_mom6 as rm6
 import xarray as xr
 from CrocoDash import logging
 from CrocoDash.extract_forcings import utils
 from CrocoDash.grid import Grid
+from CrocoDash.topo import Topo
+from CrocoDash.raw_data_access.registry import ProductRegistry
```
alperaltuntas: This is unused. Leftover?

CrocoDash/extract_forcings/obc.py
```diff
@@ -51,6 +54,43 @@ def _make_date_pairs(start: datetime, end: datetime, step_days):
     return pairs
 
 
+def _ocean_bbox_for_boundary(hgrid, tmask, boundary: str) -> dict:
+    """Return a lat/lon bounding box covering only the ocean cells on a boundary edge.
+
+    Uses tmask (shape ny×nx) to exclude land cells, giving a tighter download bbox
+    than the full supergrid extent. Falls back to the full edge extent if all cells
```
alperaltuntas, line 60: The way this docstring describes the fall back (the full edge extent) is inconsistent with what's actually being returned:  T-center extent

CrocoDash/extract_forcings/obc.py
```diff
@@ -51,6 +54,43 @@ def _make_date_pairs(start: datetime, end: datetime, step_days):
...
+    are land.
+    """
+    # T-cell centers are at every other supergrid node starting at index 1
+    tcell_lon = hgrid.x.values[1::2, 1::2]  # (ny, nx)
```
alperaltuntas: 1::2, 1::2] gives T-cell centers, but the OBC segments live on the outer supergrid line. rm6 builds the regrid target from hgrid.y.isel(nyp=0), and MOM6 reads every segment field off the Q-point line. So the box we request is half a cell off from the points we interpolate onto: for a south boundary on a 0.1 deg grid we ask for lat_min = lat_max = 7.05 while the segment sits at 7.00.

Because the accessors pad the request (+/-1 in the GLORYS functions, +/-1.5 in mom6_output.py), this isn't an issue unless the cells are coarser than ~2 degrees, but we should still fix it. So can we bound the corners of the selected ocean cells rather than their centers, i.e., x[2j:2j+3:2, 2i:2i+3:2] for cell (j,i)? Same for the all-land fallback below.

CrocoDash/extract_forcings/obc.py
```diff
@@ -403,7 +476,7 @@ def process_obc_conditions(
             start_date=start_date,
             end_date=end_date,
             get_step_days=get_step_days,
-            hgrid_path=str(hgrid_path),
+            latlon=boundary_bboxes[boundary],
             output_dir=str(raw_path),
             product_name=product_name,
             function_name=function_name,
```
alperaltuntas: Why don't we scope the context manager to the copernicusmarine accessor only? I am asking because, currently, this wraps the whole GET loop. get_glorys_data_from_rda uses xr.open_mfdataset(..., parallel=True) over one file per day and then calls to_netcdf, so the synchronous scheduler serializes that compute too. For a multi-year fetch from RDA that's probably a sizeable slowdown. Keying it off function_name would work too.

CrocoDash/extract_forcings/obc.py
```diff
@@ -168,25 +208,29 @@ def _get_boundary(
...
-            output_filename=output_filename,
-            variables=variables,
-            extra_args=extra_args,
-        )
+    # copernicusmarine opens S3-backed zarr and calls dask.compute() internally
+    # during to_netcdf(). Without this, that compute() routes to the distributed
+    # scheduler, which tries to serialize botocore.client.S3 across processes and
+    # fails. synchronous keeps it in-process. The outer parallelism (one worker
+    # per boundary/chunk) is unaffected.
+    with dask.config.set(scheduler="synchronous"):
```
alperaltuntas, line 233: Why don't we scope the context manager to the copernicusmarine accessor only? I am asking because, currently, this wraps the whole GET loop. get_glorys_data_from_rda uses xr.open_mfdataset(..., parallel=True) over one file per day, and then calls to_netcdf, so the synchronous scheduler serializes that too. For a multi-year fetch from RDA that's a big slowdown.

CrocoDash/extract_forcings/obc.py
```diff
@@ -384,6 +434,29 @@ def process_obc_conditions(
 
     variables, extra_args = utils.build_forcing_request(product_info, function_args)
 
+    # Compute per-boundary download bboxes using the bathymetry tmask so we only
+    # request data over ocean cells (tighter than the full supergrid edge extent).
+    hgrid_ds = xr.open_dataset(hgrid_path)
```
alperaltuntas: Can we open hgrid_ds with a `with` statement, so it doesn't remain open indefinitely? Also, Grid.get_bounding_boxes has an assert not Grid.is_cyclic_x(hgrid) guard, and the tmask branch skips that call entirely, so a cyclic-x grid no longer errors out, meaning we'd silently compute a non-wrapping lon min/max across the seam.

CrocoDash/extract_forcings/obc.py
```diff
@@ -384,6 +434,29 @@ def process_obc_conditions(
...
+    if bathymetry_path:
+        grid_obj = Grid.from_supergrid(hgrid_path)
+        with xr.open_dataset(bathymetry_path) as bds:
+            min_depth = bds.attrs.get("min_depth")
```
alperaltuntas: This returns None when the topog file doesn't carry the attribute, and passing min_depth=None explicitly overrides the 0.0 default in from_topo_file, which probably leads to a TypeError down the line. Does it make sense to change this as `.get("min_depth", 0.0)`?

CrocoDash/extract_forcings/obc.py
```diff
@@ -384,6 +434,29 @@ def process_obc_conditions(
...
+        boundary_bboxes = {
+            b: _ocean_bbox_for_boundary(hgrid_ds, topo.tmask, b) for b in boundaries
+        }
+        logger.info("Using tmask-derived bounding boxes for OBC data download.")
+    else:
+        full_bboxes = Grid.get_bounding_boxes(hgrid_ds)
+        boundary_bboxes = {b: full_bboxes[b] for b in boundaries}
+        logger.info("No bathymetry_path given; using full supergrid bounding boxes.")
+
+    fill_method = rm6.regridding.fill_missing_data
```
alperaltuntas, line 478: This duplicates line 464. please remove.

alperaltuntas (changes requested):

CrocoDash/extract_forcings/obc.py
manishvenu: Removed

CrocoDash/extract_forcings/obc.py
manishvenu: Replaced

CrocoDash/extract_forcings/obc.py
manishvenu: Enrico is doing parallelization work, so I'm going to pass on this comment for now.

CrocoDash/extract_forcings/obc.py
manishvenu: Done

CrocoDash/extract_forcings/obc.py
manishvenu: Yes

CrocoDash/extract_forcings/obc.py
manishvenu: Done

CrocoDash/extract_forcings/obc.py
manishvenu: replaced tmask with supergridmask

alperaltuntas (approved):

PR #227, opened 2026-06-01
Check for accidental submodule bumps & reset some submodule issues

manishvenu:
Add a check to make sure we don't merge in bad submodule updates (like the shareable config PR). You can change the submodule if you add the label "submodule-bump" which is clearly stated in the error message.

### DIFF
--- a/.github/workflows/check_submodule_bumps.yml
+++ b/.github/workflows/check_submodule_bumps.yml
@@ -24,13 +24,13 @@ jobs:
 
           git fetch origin main
 
-          CHANGED=$(git diff origin/main HEAD -- rm6 visualCaseGen)
+          CHANGED=$(git diff origin/main HEAD -- rm6 visualCaseGen demos)
 
           if [ -n "$CHANGED" ]; then
             echo "ERROR: Submodule pointer(s) changed in this PR."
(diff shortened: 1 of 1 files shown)


.github/workflows/check_submodule_bumps.yml
```diff
@@ -0,0 +1,40 @@
...
+
+          git fetch origin main
+
+          CHANGED=$(git diff origin/main HEAD -- rm6 visualCaseGen demos)
+
+          if [ -n "$CHANGED" ]; then
+            echo "ERROR: Submodule pointer(s) changed in this PR."
+            echo ""
+            echo "Changed submodule(s):"
+            git diff origin/main HEAD -- rm6 visualCaseGen demos | grep "^[-+]Subproject" || true
```
alperaltuntas: paths here need to be corrected

alperaltuntas (changes requested):

.github/workflows/check_submodule_bumps.yml
manishvenu: Done, my bad.

alperaltuntas (approved):

PR #230, opened 2026-06-02
GLORYS change longitude slicer

### LINKED ISSUE #142: [Bug report] Incomplete initial condition generated by case.configure_forcings
Hey CrocoDash team,

I was generating a regional simulation for the Ross Sea and found that the initial condition generated by case.configure_forcings() contains missing values (see the plot below). 
Image

The simulation can still run, and the missing values are probably automatically filled through interpolation from nearby grid points. The first model output looks like:
Image

The notebook used to generate this simulation:
/glade/work/hwei/crocodile_2025/CrocoDash/Project/ROSS_Worked.ipynb

The notebook used to create animations of the model output (and the plots above):
/glade/work/hwei/crocodile_2025/CrocoDash/Project/Sea_ice_animation_worked.ipynb

The initial condition files produced by configure_forcings() and process_forcings() are located at:
/glade/work/hwei/crocodile_2025/croc_input/ROSS_JAS/glorys/large_data_workflow/regridded_data

The simulation folder is:
/glade/work/hwei/crocodile_2025/croc_cases/ROSS_JAS

Please let me know if you need additional information to reproduce the issue!

Thanks,
Huaiyu

manishvenu:
Fixes #142


CrocoDash/raw_data_access/datasets/glorys.py
```diff
@@ -87,30 +85,11 @@ def get_glorys_data_from_rda(
...
-
-            # convert longitude from degree west to degree east
-            dataset["longitude"] = (360 - dataset["longitude"]) % 360
-            dataset = dataset.sortby("longitude")
+        buf = 1.0  # buffer in degrees to ensure we have enough data for interpolation at the boundaries
 
+        dataset = longitude_slicer(
+            ds, [lon_min - buf, lon_max + buf], longitude_coords="longitude"
+        )
         dataset.to_netcdf(path)
```
alperaltuntas, line 94: It looks like the latitude slicing got dropped along with the old longitude logic. The old code selected latitude between lat_min - 1 and lat_max + 1 in both branches, but the new code passes the full dataset to longitude_slicer, which only slices in longitude. As a result, the file written out now contains all global latitudes, which makes it several times larger than necessary and slows down the netcdf write. I suggest adding a latitude selection before the longitude_slicer call, e.g., ds = ds.sel(latitude=slice(lat_min - buf, lat_max + buf)). Note that the existing slow test for this function asserts that the first and last latitude values are within one degree of lat_min and lat_max, so it should catch this if run on glade.

CrocoDash/raw_data_access/datasets/glorys.py
```diff
@@ -75,9 +76,6 @@ def get_glorys_data_from_rda(
         ds_in_files = []
         date_strings = [date.strftime("%Y%m%d") for date in dates]
 
-        # Adjust lat lon inputs to make sure they are in the correct range of -180 to 180
-        lon_min, lon_max = convert_lons_to_180_range(lon_min, lon_max)
```
alperaltuntas, line 83: With this call removed, the convert_lons_to_180_range import at the top of the file is no longer used and can be removed as well.

alperaltuntas, line 83: Also, one thing we lose by removing this normalization is a consistent input convention. longitude_slicer expects the extent in increasing order, west then east. If a user provides a dateline crossing domain as, say, 170 and -170, the extent becomes decreasing and the slicer would silently select the wrong region. Could we add a guard for this case, e.g., add 360 to lon_max when it is smaller than lon_min, or at least an assertion with a clear error message?

CrocoDash/raw_data_access/datasets/glorys.py
```diff
@@ -87,30 +85,11 @@ def get_glorys_data_from_rda(
...
-                        **{"longitude": slice(-180, lon_max + 1)},
-                    ),
-                ],
-                dim="longitude",
-            )
-
-            # convert longitude from degree west to degree east
-            dataset["longitude"] = (360 - dataset["longitude"]) % 360
-            dataset = dataset.sortby("longitude")
+        buf = 1.0  # buffer in degrees to ensure we have enough data for interpolation at the boundaries
```
alperaltuntas: Minor point, but since the same one degree buffer should also apply to the latitude selection mentioned above, I suggest defining buf before both selections and using it in both places.

alperaltuntas (commented):
Since the bug being fixed here is the dateline crossing case, could we add a small unit test that calls longitude_slicer on a synthetic global dataset with a Ross Sea like extent, for example 160 to 210 degrees east? That would lock in the fix without requiring RDA access, since the existing rda test is slow, glade only, and does not cross the dateline.

alperaltuntas (changes requested):

CrocoDash/raw_data_access/datasets/glorys.py
manishvenu: We're running into this problem a bunch for grid creation, I'll think about this.

manishvenu: Done

CrocoDash/raw_data_access/datasets/glorys.py
manishvenu: Done

CrocoDash/raw_data_access/datasets/glorys.py
manishvenu: Done

CrocoDash/raw_data_access/datasets/glorys.py
manishvenu: Can I leave this for now, and then consider this later in a new PR? I can make an issue for it.

alperaltuntas: @manishvenu Sounds good! please open an issue.

alperaltuntas (approved):

PR #233, opened 2026-06-03
Auto-select PECOUNT from active ocean points at case init

### LINKED ISSUE #162: NTASKS layout is just wrong.
Need to take a look at the NTASKS layout, which is more of a VCG thing.

manishvenu:
**Descrip:**

Auto-selects the CIME pecount tier at case init from the number of active ocean cells, instead of leaving it at the default. Closes #162.

**Changes:**

1. Adds `_compute_pecount()` to `Case`, which counts active ocean cells from `ocn_topo.tmask` and selects a pecount tier (XS/S/M/L).

2. Uses pts/core benchmarks derived from tx2_3v3 — 170 pts/core without MARBL, 60 with — in increments of full Derecho nodes (128 cores).

3. Sets `cvars["PECOUNT"]` in `_configure_launch()` so `create_newcase` receives the appropriate `--pecount` flag.

**Minor changes:**

- Blocked on https://github.com/ESMCI/visualCaseGen/pull/41, which must merge first.


CrocoDash/case.py
```diff
@@ -860,6 +861,29 @@ def _configure_custom_runoff_grid(self, rof_grid_name):
                 "skip"  # to be generated later in process_forcings
             )
 
+    def _compute_pecount(self):
```
alperaltuntas, line 854: Have you confirmed that these pecount tiers actually select different PE layouts for our compsets? I think all the entries in the MOM config_pes.xml have pesize set to any, which means S, M, and L would all resolve to the same layout and this computation would have no effect. If that is the case, we might be better off passing an integer instead, since CIME sets NTASKS for all components directly when pecount is a plain number. Something like nodes_needed times TASKS_PER_NODE would make the computed node count actually take effect.

CrocoDash/case.py
```diff
@@ -860,6 +861,29 @@ def _configure_custom_runoff_grid(self, rof_grid_name):
...
+        """Compute the CIME pecount tier based on active ocean points and compset.
+
+        Uses target pts/core ratios derived from tx2_3v3 benchmarks:
+          - without MARBL: ~170 pts/core
+          - with MARBL:    ~60  pts/core
+
+        MOM6 is allocated full Derecho nodes (128 cores each). The tier string
+        maps nodes_needed to CIME --pecount keyword.
+        """
+        TASKS_PER_NODE = 128
```
alperaltuntas, line 864: This hardcodes Derecho, but Case accepts an arbitrary machine argument. Could we pull tasks per node from the machine config instead? See cime_interface.py in visualCaseGen.

CrocoDash/case.py
```diff
@@ -860,6 +861,29 @@ def _configure_custom_runoff_grid(self, rof_grid_name):
...
+        pts_per_core = 60 if self.bgc_in_compset else 170
+        ocean_pts = int(self.ocn_topo.tmask.sum())
+        nodes_needed = math.ceil(ocean_pts / (pts_per_core * TASKS_PER_NODE))
+        if nodes_needed < 1:
+            return "XS"
```
alperaltuntas, line 869: I don't think XS is a valid pecount value.

CrocoDash/case.py
```diff
@@ -871,6 +895,7 @@ def _configure_launch(self):
 
         # Variables that are not included in a stage:
         cvars["NINST"].value = self.ninst
+        cvars["PECOUNT"].value = self._compute_pecount()
```
alperaltuntas, line 888: When the user passes ntasks_ocn explicitly, the auto computed pecount still drives create_newcase and NTASKS_OCN gets overridden afterwards via xmlchange. That works, but it means the rest of the layout comes from the auto tier even though the user asserted a task count. Should we skip the auto pecount, or derive it from ntasks_ocn, when the override is given?

alperaltuntas (changes requested):
In addition to the below line comments, a general comment about the submodule bumps in this PR: all three appear stale and should be dropped.

CrocoDash/case.py
manishvenu: I am adding it to the MOM_interface PR!

CrocoDash/case.py
manishvenu: Ahh shoot yes will do.

CrocoDash/case.py
manishvenu: added it*

CrocoDash/case.py
manishvenu: For now...I think we don't do anything? I'm not sure we know enough to do anything about other component pe counts based on ntasks_ocn?

CrocoDash/case.py
manishvenu: The MOM_interface PR includes new sizes

CrocoDash/case.py
alperaltuntas: update?

CrocoDash/case.py
alperaltuntas: Could you point me to where XS is added?

CrocoDash/case.py
alperaltuntas: How about just skipping the auto pecount when `ntasks_ocn` is passed? That leaves CIME's default layout in place rather than half-overriding a tier.

alperaltuntas:
@manishvenu Please see my responses in above unresolved threads.

CrocoDash/case.py
manishvenu: https://github.com/CROCODILE-CESM/MOM_interface/pull/10. Didn't realize I hadn't merged this in.

CrocoDash/case.py
manishvenu: So this is in the PR: https://github.com/CROCODILE-CESM/MOM_interface/pull/10

CrocoDash/case.py
manishvenu: Done

CrocoDash/case.py
manishvenu: Ok I did it, but it's a little clunky because I had to add new default args  (ntasks_ocn_specified) to two funcs.

manishvenu:
@alperaltuntas Everything should be responded to now

alperaltuntas (approved):

PR #236, opened 2026-06-09
Fix the hanging issue in merge_piecewise_dataset

manishvenu:
Try fixing merge_piecewise threading issue


CrocoDash/extract_forcings/merge_piecewise_dataset.py
```diff
@@ -103,6 +104,8 @@ def merge_piecewise_dataset(
                     combine="nested",
                     concat_dim="time",
                     coords="minimal",
+                    parallel=False,
+                    lock=threading.Lock(),
```
alperaltuntas, line 111: Could you describe why this lock is needed. Also, have you checked if this work with dask? A test or two would be nice.

manishvenu: Claude suggested, but basically sometimes the open_mfdataset hangs (it's hard to reproduce why) so know I added a couple args Claude suggested to make hanging less likely. It's pretty clear it's a threading thing, so hopeful this Lock (which means only worker can read at a time) will stop those hanging issues. I've also added a test that on ocassion hangs in a similar places when we don't have this change.

PR #224, opened 2026-05-28
Add PBS batch script for driver.py

manishvenu:
Details: adds `driver.pbs` — a PBS batch submission template for running `driver.py` on HPC systems like Derecho — alongside a `case_setup/README.md` with usage docs for both interactive and batch workflows.

Closes #135

PR #237, opened 2026-06-10
Add crocodash template CLI command

### LINKED ISSUE #135: [Feature Request] A too_much_data pbs script (plus an example of this)
The `too_much_data` function really is quite fast. However, running `driver.py` in the terminal can be stressful (for me) because sometimes my session would time out before all the boundary conditions were finished downloading/processed. This could be fixed (I think) with a `driver.pbs` script that would allow `driver.py` to run in the background and independently of the Jupyter server. 

Below is an implementation of a `large_data_workflow/driver.pbs` that worked for me. I just executed it with qsub: 

```
#!/bin/bash -e
# Batch system directives
#PBS -N python_job
#PBS -r n
#PBS -j oe
#PBS -S /bin/bash
#PBS -l select=1:ncpus=1:mpiprocs=1:ompthreads=1:mem=50GB
#PBS -A CESM0030
#PBS -q main
#PBS -l walltime=06:00:00

source ~/.bash_profile
module load conda
eval "$(mamba shell hook --shell bash)"
conda activate CrocoDash

python driver.py

```

manishvenu:
## Summary

- Adds `crocodash template` subcommand to generate a starter CrocoDash case file, or a PBS submission script
- `--kind case` (default): output can be a YAML config (`.yaml`/`.yml`), Jupyter notebook (`.ipynb`), or Python script (`.py`) — the `.py` variant extracts code cells from the gallery tutorial notebook, so there is no separate template file to maintain
- `--kind pbs`: writes a PBS batch script that runs `crocodash process --caseroot <caseroot> --all` on an HPC queue (e.g. Derecho) instead of interactively. Supersedes #224 — the PBS script itself now lives in CrocoGallery instead of being carried in CrocoDash
- `--machine derecho` (or any key in `demos/tools/known_paths.json`) pre-fills `<KEY>` placeholders with real dataset paths; omitting `--machine` leaves placeholders intact for manual editing
- Passes a helpful error listing available machine names if an unknown `--machine` value is given
- Depends on the refactored `inject_paths.py` in CrocoGallery (see https://github.com/CROCODILE-CESM/CrocoGallery/pull/38)

Closes #135

## Usage

```bash
# Notebook with placeholders
crocodash template --output my_case.ipynb

# Notebook with Derecho paths pre-filled
crocodash template --output my_case.ipynb --machine derecho

# Python script with Derecho paths pre-filled
crocodash template --output my_case.py --machine derecho

# YAML config with Derecho paths pre-filled
crocodash template --output my_case.yaml --machine derecho

# PBS batch script for submitting forcing extraction to a queue
crocodash template --output submit_forcings.pbs --kind pbs --machine derecho
```


environment.yml
```diff
@@ -21,6 +21,7 @@ dependencies:
     - -e ./CrocoDash/rm6/
     - -e ./CrocoDash/visualCaseGen/external/ipyfilechooser/
     - -e ./CrocoDash/visualCaseGen/
+    - -e ./demos/
```
alperaltuntas, line 25: Why is demos a package now?

CrocoDash/cli.py
```diff
@@ -45,6 +45,41 @@ def _duplicate_case(args):
     print(f"Duplicated case created at: {new_case.caseroot}")
 
 
+def _template(args):
```
alperaltuntas, line 132: When I pass an unknown machine or notebook name, or an output path whose parent directory doesn't exist, I get a full Python traceback instead of a clean error message. And since KeyError reprs its argument, the newlines in the notebook listing are printed as literal backslash-n, so the "available notebooks" list comes out as one long unreadable line. Could you catch KeyError and FileNotFoundError here, print the message, and exit with a nonzero code? That would also make the docs example (which currently shows a raw KeyError as the expected output) look intentional.

CrocoDash/cli.py
```diff
@@ -45,6 +45,41 @@ def _duplicate_case(args):
...
+    from pathlib import Path
+    from crocogallery import get_notebook_path, inject_into_text, load_paths
+
+    output = Path(args.output)
+    notebook_id = args.notebook
+
+    if output.suffix in (".yaml", ".yml"):
+        # YAML starter lives alongside the default tutorial notebook
+        yaml_path = get_notebook_path(notebook_id).parent / "starter_case.yaml"
```
alperaltuntas: This resolves `starter_case.yaml` relative to whichever notebook is passed via `--notebook`, but only the tutorials directory has that file. So `--output x.yaml --notebook crocodash.features.add_bgc` fails with a FileNotFoundError pointing at an internal gallery path. I tried this locally and confirmed it. Since the YAML starter has nothing to do with the chosen notebook, I'd resolve it from the default tutorial directory unconditionally, or raise a clear error when `--notebook` is combined with a yaml output.

CrocoDash/cli.py
```diff
@@ -45,6 +45,41 @@ def _duplicate_case(args):
...
+        template_text = yaml_path.read_text()
+        if args.machine:
+            template_text = inject_into_text(template_text, load_paths(args.machine))
+        output.write_text(template_text)
```
alperaltuntas: `write_text` fails with a raw FileNotFoundError if the parent directory of the output path doesn't exist. Please add `output.parent.mkdir(parents=True, exist_ok=True)` before writing (same applies to the notebook and py branches below).

CrocoDash/cli.py
```diff
@@ -45,6 +45,41 @@ def _duplicate_case(args):
...
+        paths = load_paths(args.machine) if args.machine else {}
+        nb = nbformat.read(get_notebook_path(notebook_id), as_version=4)
+
+        if output.suffix == ".ipynb":
+            for cell in nb.cells:
+                if cell.cell_type == "code":
+                    cell.source = inject_into_text(cell.source, paths)
+            nbformat.write(nb, output)
+        else:
+            code_cells = [cell.source for cell in nb.cells if cell.cell_type == "code"]
```
alperaltuntas: The markdown cells are dropped entirely here, so users of the py output lose all the narration that notebook users get. What do you think about emitting them as `# %% [markdown]` comment blocks (the jupytext convention)? That would preserve the guidance text at little cost.

CrocoDash/cli.py
```diff
@@ -170,5 +205,32 @@ def main():
...
+        default=None,
+        help="Pre-fill known dataset paths for this machine (e.g. derecho). Omit to leave <KEY> placeholders.",
+    )
+    template_parser.add_argument(
+        "--notebook",
+        default="crocodash.tutorials.crocodash_tutorial",
+        help=(
+            "Gallery notebook ID to use as the template source "
+            "(default: crocodash.tutorials.crocodash_tutorial). "
+            "Run `python -c 'from crocogallery import list_notebooks; print(*list_notebooks(), sep=\"\\n\")'` "
```
alperaltuntas: Asking users to run a python one-liner to discover notebook IDs feels a bit awkward. Why not add a `--list-notebooks` flag to the template subcommand instead?

.github/scripts/check_glade_quality_control.sh
```diff
@@ -6,6 +6,7 @@ EXEMPTIONS=(
   "CrocoDash/raw_data_access/datasets/mom6_output.py"
   "tests/fixtures/state.py"
   "tests/extract_forcings/test_processes.py"
+  "tests/template/test_cli.py"
```
alperaltuntas: Is this exemption actually needed? I grepped `tests/template/test_cli.py` and it contains no `/glade` strings at all.

tests/template/test_cli.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+# --- --notebook flag ---
+
+
+def test_template_custom_notebook(tmp_path):
+    """Any gallery notebook can be used as the template source."""
+    notebooks = list_notebooks()
+    # pick a notebook other than the default
+    alt_id = next(
```
alperaltuntas, line 192: If the gallery ever ships only the default notebook, this `next` call fails with an opaque StopIteration. A `pytest.skip` in that case would make the failure mode clearer. Also, the suffix check is redundant since `list_notebooks` only returns ipynb files.

alperaltuntas (changes requested):

environment.yml
manishvenu: So we can pull notebooks from CrocoGallery through this template command which calls demos. It allows a more fluent way to inject derecho/machine paths.

manishvenu: Which will be nice for the workshop.

CrocoDash/cli.py
manishvenu: Yes will do!

.github/scripts/check_glade_quality_control.sh
manishvenu: Done

tests/template/test_cli.py
manishvenu: Done

CrocoDash/cli.py
manishvenu: Done

CrocoDash/cli.py
manishvenu: Done

CrocoDash/cli.py
manishvenu: It's resolved frm the default tutorial directory,

CrocoDash/cli.py
manishvenu: done

CrocoDash/cli.py
```diff
@@ -102,6 +104,80 @@ def _duplicate_case(args):
...
+        if output.suffix == ".ipynb":
+            for cell in nb.cells:
+                if cell.cell_type == "code":
+                    cell.source = inject_into_text(cell.source, paths)
+            nbformat.write(nb, output)
+        else:
+            blocks = []
+            for cell in nb.cells:
+                if cell.cell_type == "code":
+                    blocks.append("# %%\n" + inject_into_text(cell.source, paths))
```
alperaltuntas: The `.py` output isn't valid Python — the tutorial has a `%matplotlib ipympl` cell and it comes through verbatim, so `python my_case.py` dies with a SyntaxError on line 78. I confirmed it locally; neither the placeholder nor the `--machine derecho` version passes `py_compile`. Could you comment out magic/shell lines (`^\s*[%!]`) when emitting code cells, the way jupytext does?

CrocoDash/cli.py
```diff
@@ -275,5 +351,60 @@ def main():
...
+        dest="list_notebooks",
+        help="Print all available gallery notebook IDs and exit.",
+    )
+    template_parser.set_defaults(func=_template, subparser=template_parser)
+
     args = parser.parse_args()
-    args.func(args)
+    try:
+        args.func(args)
+    except KeyError as e:
```
alperaltuntas: This catches for every subcommand, not just `template`. I patched a bad dict lookup into `_dump` to check, and the whole output was `ocean_topog` — no exception type, no traceback, no line number. That'll be rough to debug in `create`/`process`. Could you move the try/except into `_template`, or catch a dedicated `CrocoDashCliError` here instead?

CrocoDash/cli.py
```diff
@@ -275,5 +351,60 @@ def main():
...
+    )
+    template_parser.add_argument(
+        "--output",
+        default=None,
+        help="Output path. For --kind case: .yaml, .ipynb, or .py. For --kind pbs: any path (e.g. submit_forcings.pbs). Required unless --list-notebooks is given.",
+    )
+    template_parser.add_argument(
+        "--machine",
+        default=None,
+        help="Pre-fill known dataset paths for this machine (e.g. derecho). Omit to leave <KEY> placeholders.",
```
alperaltuntas, line 420: Three of the keys in `known_paths.json` aren't real paths — `CESM` is `"Checkout"`, `inputdir` is `"fill_in_id"`, `casedir` is `"fill_in_cd"` — so `--machine derecho` gives you `cesmroot: Checkout` and `caseroot: fill_in_cd/my_case`. That's worse than leaving the placeholders, since now nothing signals that those still need editing. Can we skip those three when injecting, or fix them in CrocoGallery?

CrocoDash/cli.py
```diff
@@ -102,6 +104,80 @@ def _duplicate_case(args):
...
+        return
+
+    if not args.output:
+        args.subparser.error("--output is required unless --list-notebooks is given.")
+
+    output = Path(args.output)
+    notebook_id = args.notebook
+    output.parent.mkdir(parents=True, exist_ok=True)
+
+    if args.kind == "pbs" or output.suffix in (".yaml", ".yml"):
```
alperaltuntas: The header inside `submit_forcings.pbs` tells people to run `crocodash template --output submit_forcings.pbs --machine derecho` — no `--kind pbs` — which quietly writes the `.py` template into a file named `.pbs`. I hit this before I spotted the flag. Since `.yaml` and `.ipynb` already dispatch on suffix, why not do the same for `.pbs`? Also `--machine` is a no-op for pbs output, so the "applies to both" line in `template.md` is misleading.

alperaltuntas (changes requested):
Thanks, all eight of my earlier comments are addressed — I checked out the branch, pinned the demos submodule, and ran the command for real. Tests pass. A few things came up while trying it out.

CrocoDash/cli.py
manishvenu: Done

CrocoDash/cli.py
manishvenu: Moved to template

CrocoDash/cli.py
manishvenu: They are skipped

CrocoDash/cli.py
manishvenu: Fixed the header and you don't need machine anymore. We are dispatching on pbs now.

alperaltuntas (approved):

PR #238, opened 2026-06-10
CLI + YAML Configuration

manishvenu:
## Overview

Three related changes that clean up the forcing extraction architecture and extend the CLI.

### 1. CLI + YAML case creation (original scope)
- New `recipe.py`: create a full CrocoDash case from a YAML config file (`create_case_from_yaml`), and reconstruct a YAML from an existing case (`case_to_yaml`)
- `crocodash create --config case.yaml` — create a case from YAML
- `crocodash dump --caseroot /path/to/case` — dump an existing case to YAML stdout
- `case_state.py`: serializes case init args + derived paths to `crocodash_state.json` in the caseroot; used by recipe.py and the shareable module to reconstruct cases without a live `Case` object

### 2. Shareable module refactor
- `shareable/` collapses into a single `shareable.py` — `CaseBundle`, `ForkBundle`, `duplicate_case`
- Reads case state via `case_state.read()` instead of re-running case init
- Removed flags/args that are now handled automatically from state (`--compset`, `--extra-configs`, `--remove-configs`, `--extra-forcing-args`)

### 3. Replace case_setup/driver.py with `crocodash process` CLI
- Delete `extract_forcings/case_setup/` entirely — no more copytree or importlib loading
- New `extract_forcings/driver.py`: `run_workflow(config_path, ...)` derives all paths from `case_state.read()` instead of storing grid paths in `config.json`
- `config.json` `"basic"` key renamed to `"conditions"`; grid/path sub-section dropped; `"caseroot"` added so the driver can always locate the case state
- `Config` class in `extract_forcings/utils.py` deleted
- `crocodash process` CLI subcommand with flexible config resolution:
  - `--config PATH` — direct path to `config.json` (highest priority)
  - `--caseroot PATH` — derive config path from case state
  - auto-detect `config.json` in cwd (for running from inside `extract_forcings/`)
  - default: treat cwd as caseroot
- Per-component flags: `--all`, `--ic`, `--bc`, `--bgcic`, `--bgcironforcing`, `--bgcrivernutrients`, `--runoff`, `--tides`, `--chl`, `--skip NAME...`
- Tests rewritten in `tests/extract_forcings/test_driver.py` and new `tests/extract_forcings/test_cli.py` (19 tests)
- Docs updated: `cli.md` process section, `extract_forcings.md` rewritten around CLI/Python API, RST api-docs updated

### 4. Move OBC/IC configuration into a `ConditionsConfigurator`
- The IC/OBC `user_nl_mom` params and the `config.json` `"conditions"` entry, previously hand-built in `case.py` (`configure_initial_and_boundary_conditions` / `_update_forcing_variables`), now go through a new `ConditionsConfigurator` in `forcing_configurations/configurations.py` — the same declarative registry pattern already used for tides/BGC/CICE/runoff/chl
- `case.py` is trimmed to date_range shape validation + Case-level state assignment (`boundaries`, `date_range`, `forcing_product_name`); `configure_initial_and_boundary_conditions` and `_update_forcing_variables` are removed, with the remaining 3 xmlchanges (`RUN_STARTDATE`/`STOP_OPTION`/`STOP_N`) inlined into `configure_forcings`
- `boundaries`/`product_name` validation moved into `ConditionsConfigurator.validate_args()`
- Added a `registry` back-reference on `BaseConfigurator` so one configurator can read a sibling's helper methods — used to pull `TidesConfigurator.tidal_data_str()` into the composite `OBC_SEGMENT_*_DATA` string without duplicating `append_user_nl` writes
- Added a `ConfigOutputParam` type for derived values (dates, product metadata, boundary numbering) that only need to land in `config.json`, not in `user_nl`/xml
- `config.json`'s `"conditions"` entry now uses the generic `serialize()` `inputs`/`outputs` shape instead of the old bespoke `dates`/`forcing`/`general` structure; `driver.py`, `recipe.py`, and `cli.py` read sites updated to match
- Test fixtures in `test_driver.py`, `test_cli.py`, `test_recipe_functions.py`, `test_case_integration.py` updated to the new shape; `test_all_configurators_smoke` extended to cover boundaries/product_name and the new `ConfigOutputParam` type


CrocoDash/case.py
```diff
@@ -125,11 +146,13 @@ def __init__(
         )
 
         # Set instance attributes
```
alperaltuntas: Can we reorder this whole attribute setting block so that it follows the order of the arguments, and clearly separate those that are set by arguments from those that are not.

CrocoDash/recipe.py
```diff
@@ -0,0 +1,265 @@
+import json
```
alperaltuntas, line 10: Please add a module-level docstring describing what a recipe is, why it's needed and how it's used.

CrocoDash/case.py
```diff
@@ -872,6 +901,25 @@ def _configure_launch(self):
         # Variables that are not included in a stage:
         cvars["NINST"].value = self.ninst
 
+    def _write_state(self):
+        """Write case creation parameters to crocodash_state.json in caseroot."""
+        state = {
```
alperaltuntas: I think we need semantic versioning of state schema to determine compatibility.

alperaltuntas: Can you pull state dictionary alongside its read/write/validate/version into a dedicated class+module to declutter case.py? This should also give a single location for the serialization policy that's currently split across case.py's _SERIALIZABLE_EXCLUDE and recipe.py's _STATE_DERIVED_KEYS (the two hardcoded lists that must stay in sync.

CrocoDash/recipe.py
```diff
@@ -0,0 +1,265 @@
...
+
+    case = Case(
+        ocn_grid=grid,
+        ocn_topo=topo,
+        ocn_vgrid=vgrid,
+        override=override,
+        **config["case"],
+    )
+
+    case.configure_forcings(**config["forcings"])
```
alperaltuntas, line 162: What happens if the original case didn't run `configure_forcings`? Can a case be shared before configure_forcings? More generally, when can a case be in shareable form?

CrocoDash/shareable.py
```diff
@@ -0,0 +1,752 @@
...
+# ---------------------------------------------------------------------------
+
+
+def ask_string(prompt: str, default="") -> str:
+    try:
+        response = input(prompt).strip()
+        return response if response else default
+    except EOFError:
+        print("\nNo input detected, returning empty string.")
+        return ""
```
alperaltuntas: Why do you return an empty string instead of the default here?

CrocoDash/shareable.py
```diff
@@ -0,0 +1,752 @@
...
+            When omitted the recipient is asked interactively for each category.
+        """
+        config = self._configure_yaml_for_forked_case_args(
+            cesmroot, machine, project_number, new_caseroot, new_inputdir
+        )
+        config = self._guide_yaml_review(config)
+        self._resolve_copy_plan(plan)
+
+        logger.info("Creating new case from YAML...")
+        self.case = create_case_from_yaml(config, override=True)
```
alperaltuntas: This runs process_forcings() (needs raw source data) before the bundle/ocnice files are copied. On a machine without that raw data, fork fails before the shipped files are ever used.

CrocoDash/shareable.py
```diff
@@ -0,0 +1,752 @@
...
+            dst = Path(self.case.inputdir) / "ocnice" / src.name
+            if not dst.exists():
+                shutil.copy(src, dst)
+
+        logger.info("Applying non-standard CESM state per plan...")
+        self.apply_copy_plan()
+
+        self.case.validate_case()
+
+        print(
```
alperaltuntas: This is redundant.

CrocoDash/shareable.py
```diff
@@ -0,0 +1,752 @@
...
+            if not line.startswith("./xmlchange"):
+                continue
+            _, kv = line.split(None, 1)
+            param, value = kv.split("=", 1)
+            if param in xmlchangeparams:
+                logger.info(f"Running {param}={value} xmlchange to new caseroot")
+                xmlchange(param, value, is_non_local=True)
+
+
+def copy_configurations_to_case(old_forcing_config, case, inputdir_ocnice):
```
alperaltuntas: where is this function called? If nowhere, is it still necessary?

CrocoDash/shareable.py
```diff
@@ -0,0 +1,752 @@
...
+    from CIME.case import Case
+
+    return Case(caseroot, read_only=True, non_local=True)
+
+
+def _run_xmlquery(caseroot, param):
+    res = subprocess.run(
+        ["./xmlquery", param, "-N"], cwd=str(caseroot), capture_output=True
+    )
+    return res.stdout.decode().strip().split(":")[1].strip()
```
alperaltuntas, line 743: Rather than swallowing subprocess failures, pass the error message to the user.

alperaltuntas (changes requested):

CrocoDash/shareable.py
manishvenu: configure_only was added to this.

CrocoDash/shareable.py
manishvenu: removed print statement

CrocoDash/recipe.py
manishvenu: Done

CrocoDash/recipe.py
manishvenu: The case can't be shared. A case can be shared after configure_forcings

CrocoDash/shareable.py
manishvenu: It's legacy and was fairly okay, but it's now removed

CrocoDash/shareable.py
manishvenu: Mistake fixed

CrocoDash/case.py
manishvenu: Done

CrocoDash/case.py
manishvenu: Ok Done

manishvenu: Done

CrocoDash/shareable.py
manishvenu: Done

manishvenu:
@alperaltuntas, After responding to review comments, I also added a new CLI commaand called "process" that takes the place of the  `python driver.py --all` we were doing for the offline process forcings step. Because of that we no longer need the "case_setup" folder in the extract_forcings module. We directly write the config.json from case.py and the driver stays as part of CrocoDash.


This is a pretty nice simplification that brings the old "CLI" type of thing we are doing into the official CLI.

manishvenu:
@alperaltuntas The other major simplification is that case.py used to manually do the obc & ic configuration, which was messy and not part of the forcing config registry. I moved that into the forcing_configurations folder (now that we can) sas the ConditionsConfigurator. Which is really nice and takes a couple hundred lines out of case.py

CrocoDash/shareable.py
alperaltuntas: @manishvenu It was added to `create_case_from_yaml` and wired into `duplicate_case()` and `identify_non_standard_case_info()`, but not into `fork()` itself, which is the path this comment is about.

CrocoDash/cli.py
```diff
@@ -30,40 +103,99 @@ def _duplicate_case(args):
...
+        "--bc", action="store_true", help="Run boundary conditions"
+    )
+    ef_components.add_argument(
+        "--bgcic", action="store_true", help="Run BGC initial conditions"
+    )
+    ef_components.add_argument(
+        "--bgcironforcing", action="store_true", help="Run BGC iron forcing"
+    )
+    ef_components.add_argument(
+        "--bgcrivernutrients", action="store_true", help="Run BGC river nutrients"
```
alperaltuntas, line 182: Do we need the user to specify `--runoff` when `--bgcrivernutrients` is specified. If so, could you add a check?

manishvenu: Now runoff is default enabled if bgcrivernutrients is enabled. (Changed in driver.py)

CrocoDash/shareable.py
manishvenu: I think this will work! Adding configure only and copying over the configuration files  is fine. Then validate case will let the user know to run more extract forcings path if the config changed.

CrocoDash/extract_forcings/utils.py
```diff
@@ -1,11 +1,11 @@
...
 import re
 from pathlib import Path
 from collections import defaultdict
 from datetime import datetime, timedelta
 from CrocoDash.topo import *
 from CrocoDash.grid import *
+import json
 
 
 class Config:
```
alperaltuntas: The merge commit brought this removed class back, which I believe is an inadvartent merge artifact. Please investigate the code to see if any other undesired merge artifacts got introduced.

alperaltuntas (changes requested):

CrocoDash/extract_forcings/utils.py
manishvenu: I think they are resolved

alperaltuntas (approved):

amrhein:
🥳

PR #252, opened 2026-07-08
Enable WW3 compsets

alperaltuntas:
Enables CrocoDash to configure WW3 active wave coupling on custom/regional ocean grids, reusing the custom ocean grid as the wave grid. Previously CrocoDash required SWAV and asserted out on anything else; this PR lets a compset carry WW3 and drives the new visualCaseGen Wave-Grid stages end-to-end.

Changes:

CrocoDash/case.py
- Allow active waves: wave guard changed from "must be SWAV" to "must not be DWAV"; WW3 and SWAV accepted, DWAV still unsupported.
- New ww3_in_compset property.
- WW3 grid-input generation in _create_grid_input_files() via ocn_topo.write_ww3_input(ocnice, grid_alias=ocn_grid.name) when WW3 is present (also tidied ocnice path building with a shared suffix).
- New _configure_custom_wave_grid() (called from _configure_custom_grid()): sets WAV_GRID_MODE="Custom Ocean Grid" and WW3_INPUT_STATUS="Complete" when the Wave Grid Mode stage is active; no-op for SWAV.
- Minor validate_case() cleanup.

Future direction:
- add WW3 OBC feature.

**WARNING**: This PR breaks backwards compatibility: From now on, we require cesm3_0_alpha09d or newer.


manishvenu:
Do we need to update the CESM version in the container? Otherwise this looks good.

manishvenu (approved):
Approving Alper's WW3 Compsets, then adding in the BGC changes and new CESM.

manishvenu:
@alperaltuntas I have a succesful run with WW3.

PR #253, opened 2026-07-13
Fix OBC/GLORYS antimeridian bugs found building polar test domains

manishvenu:
## Summary
Two related bugs in the OBC forcing pipeline, found while building pole-centered Arctic/Antarctic test domains for a location-dependent MOM6/CESM scaling comparison (any domain that fully encircles a pole necessarily has edges that cross the antimeridian and/or the prime meridian):

- **`extract_forcings/obc.py`**: `_regrid_boundary` sliced each `regrid_step`-sized chunk with `slice(chunk_start, chunk_end)`, where both bounds are midnight-anchored dates. Daily-mean products like GLORYS timestamp each day's value at noon, so the slice's upper bound excluded `chunk_end`'s own data point. This silently dropped the last day of every multi-day chunk, and produced a **fully empty selection** for a single-day trailing chunk (e.g. a date range that isn't an exact multiple of `regrid_step_days`), crashing downstream with a zero-size-array `IndexError`.
- **`raw_data_access/datasets/glorys.py`**: `get_glorys_data_from_rda` routed any opposite-sign `(lon_min, lon_max)` pair into its antimeridian-split-and-concat branch, but that's also true for a box straddling the prime meridian (e.g. `-45..45`), not just genuine antimeridian wraparound (e.g. `170..-170`). For the former, the two slices overlap heavily instead of being disjoint, **duplicating a large swath of longitude values** in the concatenated result — silently tolerated by `xesmf`'s regridder (`ignore_degenerate=True`) but not by `xarray`'s `interpolate_na` (used for initial-condition gap-filling), which crashed with "Index 'longitude' has duplicate values".

Also includes an incidental submodule pointer update for `rm6`, pointing at the fixed `regional-mom6` commit (degenerate-cell OBC regrid fix, see CROCODILE-CESM/regional-mom6#68).

## Test plan
- [x] Verified the exact `IndexError`/zero-size-chunk crash no longer occurs against real GLORYS raw data (before/after comparison of the chunk selection).
- [x] Verified the duplicate-longitude bug no longer occurs for both a prime-meridian-straddling box and the full-wraparound case (0 duplicates in both, previously ~50%).
- [x] Ran full Equatorial Pacific and Antarctic OBC+IC forcing pipelines end-to-end on Derecho with real GLORYS data; both now complete successfully where they previously crashed.
- [x] Built and ran CESM cases against the corrected forcings for both locations (clean runs, no FATAL/CFL/NaN errors, `read_cesm_timing` timing collected).

### DIFF
--- a/CrocoDash/extract_forcings/obc.py
+++ b/CrocoDash/extract_forcings/obc.py
@@ -248,13 +248,16 @@ def _regrid_boundary(
             continue
 
         tmp_file = output_folder / f"_tmp_{boundary}_{start_str}_{end_str}.nc"
-        # Raw product timestamps (e.g. GLORYS daily means) are stamped at
-        # noon, not midnight — push chunk_end to end-of-day so label-based
-        # .sel() doesn't silently drop the last day's sample at each chunk
-        # boundary. Mirrors make_dates_end_inclusive in raw_data_access.
-        chunk_end_inclusive = chunk_end + timedelta(hours=23, minutes=59, seconds=59)
-        end_str = chunk_end_inclusive.strftime("%Y-%m-%d")
-        ds_full.sel(time=slice(start_str, end_str)).to_netcdf(tmp_file)
+        # chunk_start/chunk_end are midnight-anchored dates, but daily-mean products
+        # like GLORYS timestamp each day's value at noon. A plain slice(chunk_start,
+        # chunk_end) therefore excludes chunk_end's own data point (it falls after
+        # the midnight upper bound) -- silently dropping the chunk's last day, and
+        # producing a fully empty selection for any single-day chunk (e.g. the
+        # trailing remainder when the date range isn't a multiple of
+        # regrid_step_days), which crashes downstream on a zero-size array. Extend
+        # the upper bound through the end of chunk_end's calendar day.
+        chunk_end_of_day = chunk_end + timedelta(hours=23, minutes=59, seconds=59)
+        ds_full.sel(time=slice(chunk_start, chunk_end_of_day)).to_netcdf(tmp_file)
 
         try:
             seg = rm6.segment(
--- a/CrocoDash/extract_forcings/obc.py
+++ b/CrocoDash/extract_forcings/obc.py
@@ -248,16 +248,11 @@ def _regrid_boundary(
             continue
 
         tmp_file = output_folder / f"_tmp_{boundary}_{start_str}_{end_str}.nc"
-        # chunk_start/chunk_end are midnight-anchored dates, but daily-mean products
-        # like GLORYS timestamp each day's value at noon. A plain slice(chunk_start,
-        # chunk_end) therefore excludes chunk_end's own data point (it falls after
-        # the midnight upper bound) -- silently dropping the chunk's last day, and
-        # producing a fully empty selection for any single-day chunk (e.g. the
-        # trailing remainder when the date range isn't a multiple of
-        # regrid_step_days), which crashes downstream on a zero-size array. Extend
-        # the upper bound through the end of chunk_end's calendar day.
-        chunk_end_of_day = chunk_end + timedelta(hours=23, minutes=59, seconds=59)
-        ds_full.sel(time=slice(chunk_start, chunk_end_of_day)).to_netcdf(tmp_file)
+        # Daily-mean products like GLORYS timestamp each day's value at noon, so
+        # slice by date strings (pandas partial-string indexing treats the end
+        # string as covering that whole calendar day) to include chunk_end's own
+        # data point instead of a midnight-anchored datetime slice excluding it.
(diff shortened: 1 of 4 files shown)


CrocoDash/raw_data_access/datasets/glorys.py
```diff
@@ -89,7 +89,18 @@ def get_glorys_data_from_rda(
...
+        # that happens to straddle the prime meridian (e.g. -45..45) -- so a direct
+        # slice is correct either way. Splitting is only needed when the range
+        # actually wraps around the antimeridian (lon_min > lon_max, e.g. 170..-170).
+        # The previous `lon_min * lon_max > 0` (same-sign) check wrongly routed any
+        # opposite-sign range into the split/concat branch too, including
+        # prime-meridian-straddling boxes -- there, the two slices
+        # slice(lon_min-1, 360) and slice(-180, lon_max+1) overlap heavily (both
+        # cover the whole middle region) instead of being disjoint, duplicating a
+        # large swath of longitude values in the concatenated result.
+        if lon_min <= lon_max:
```
alperaltuntas, line 111: This check is right and also fixes the case where lon_min or lon_max is exactly zero, which the old product test misrouted as well. Two residual corner cases worth considering: First, the one degree buffers in the wrap branch mean a near full circle box, say lon_min 0.5 and lon_max -0.5, still produces overlapping slices and duplicate longitudes, the same failure class this PR fixes at a narrower trigger. Second, convert_lons_to_180_range maps exactly 180 to -180, so a box like 100 to 180 arrives as 100 to -180 and takes the wrap branch unnecessarily.

CrocoDash/extract_forcings/obc.py
```diff
@@ -260,7 +260,16 @@ def _regrid_boundary(
...
-        ds_full.sel(time=slice(chunk_start, chunk_end)).to_netcdf(tmp_file)
+        # chunk_start/chunk_end are midnight-anchored dates, but daily-mean products
+        # like GLORYS timestamp each day's value at noon. A plain slice(chunk_start,
+        # chunk_end) therefore excludes chunk_end's own data point (it falls after
+        # the midnight upper bound) -- silently dropping the chunk's last day, and
+        # producing a fully empty selection for any single-day chunk (e.g. the
+        # trailing remainder when the date range isn't a multiple of
+        # regrid_step_days), which crashes downstream on a zero-size array. Extend
+        # the upper bound through the end of chunk_end's calendar day.
+        chunk_end_of_day = chunk_end + timedelta(hours=23, minutes=59, seconds=59)
```
alperaltuntas: This resolves the day slipping between chunks that I asked about on PR 218, and I confirmed the extension cannot double count a midnight stamped point since the next chunk starts at the following midnight. That said, there is a simpler idiom that avoids the 23:59:59 arithmetic entirely: pass the date strings you already build two lines above, ds_full.sel(time=slice(start_str, end_str)). Pandas partial string indexing treats the end string as covering that whole calendar day, which is exactly the behavior you want. That would also let this comment shrink to a couple of lines stating the constraint, something like daily products stamp values at noon, so slice by date strings to include the full end day. The current ten line block reads more like the PR description than code documentation.

CrocoDash/raw_data_access/datasets/glorys.py
alperaltuntas: The routing fix above is correct, but I think the relabel at line 124 here, in the wrap branch, has a latent bug that your antimeridian domains will now exercise. The relabel (360 - lon) % 360 is a reflection, not a shift. A point at 169 east maps to 191, and a point at 169 west, stored as -169, maps to 169, so the two sides of the dateline swap places. The correct conversion to the 0 to 360 convention is lon % 360, which sends -169 to 191 and leaves 169 alone. The comment says degree west, but the input coordinate is the standard east positive -180 to 180 range, the same one sliced two lines up. Note that your zero duplicates check cannot catch this, since a mirror produces no duplicates, and a CESM run with mirrored boundary forcing can still complete cleanly. Since the Equatorial Pacific test is presumably the one domain that goes through this branch, could you plot the selected field across the dateline against the raw data before and after changing this to lon % 360, and include the fix in this PR?

alperaltuntas (changes requested):

CrocoDash/extract_forcings/obc.py
manishvenu: Done

CrocoDash/raw_data_access/datasets/glorys.py
manishvenu: Done, will send plots soon

CrocoDash/raw_data_access/datasets/glorys.py
manishvenu: Done

alperaltuntas (commented):
Thanks for the fixes. Once the merge conflict(s) are resolved and CI tests pass, feel free the merge this PR.

PR #256, opened 2026-07-20
Drive OBC/tide segments directly via regional_mom6.Segment (interior boundary support)

manishvenu:
## Summary
- Points `CrocoDash/rm6` at the standalone `Segment` class (regional-mom6 `manish/standalone-boundary-class`, PR #315 on COSIMA/regional-mom6), which supports arbitrary straight OBC segments (full edge, partial edge, or fully interior line) via `Segment.from_hgrid`/`Segment.cardinal`/`Segment.from_lonlat` -- not just the 4 cardinal edges.
- Stops using `regional_mom6.experiment` for boundary/tide setup entirely -- `Experiment`'s wrapper is cardinal-only and capped at 4 segments, and isn't being extended for this; `case.py`, `extract_forcings/obc.py`, and `extract_forcings/tides.py` now drive `Segment` directly instead.
- No CrocoDash-side spec class needed: `Segment` itself now owns `from_lonlat` (define a boundary from physical lon/lat instead of raw supergrid indices) and `to_spec()`/`from_spec()` (serialize a segment's construction to a plain dict and rebuild it later against the same hgrid). A custom (non-cardinal) boundary is just a `Segment` built directly via `Segment.from_lonlat`/`from_hgrid` and passed into `Case`'s `boundaries` list alongside cardinal strings -- no `CustomSegment`/`segment_spec.py` wrapper.
- `boundary_key()`/`build_segment()` (dispatch cardinal string vs. `Segment` vs. the `custom_segments` dict read back from config.json) now live in `extract_forcings/obc.py`, shared by `obc.py` and `tides.py`.
- `boundaries` passed to `configure_forcings`/`configure_initial_and_boundary_conditions` may mix cardinal strings and `Segment` instances; `OBC_SEGMENT_00N` position strings are generated via `Segment.mom6_obc_position_string()` instead of a hardcoded cardinal `if/elif` block.
- Removed the dead `Case.expt` property (unused anywhere in the real pipeline, and broken by `Experiment`'s new read-only `hgrid`/`vgrid` properties on the new branch).
- Fixed `extract_forcings/initial_condition.py` for the same `Experiment` API change (`m6f_hgrid` instead of a settable `hgrid` property) -- unrelated to boundaries, but broken by the regional-mom6 branch bump.
- Updated `test_process_tides` to mock `Segment.regrid_tides` instead of the removed `Experiment.setup_boundary_tides`; dropped `test_case_expt_smoke` (tested the removed `expt` property).

## Test plan
- [x] Full CrocoDash pytest suite passes (`pytest --ignore=CrocoDash/visualCaseGen -m "not workflow"`: 181 passed, 14 skipped).
- [x] New regional-mom6 pytest coverage for a multi-segment domain (7 segments incl. interior/partial, overlap, order-independence) passes on the companion branch.
- [x] `Segment.from_lonlat`/`to_spec`/`from_spec` verified directly against the sketch-domain fixture: `from_lonlat` resolves to the same `OBC_SEGMENT` position string as constructing the equivalent segment from raw supergrid indices, and `to_spec`/`from_spec` round-trips exactly (position string and mask both match).
- [x] End-to-end proof: built a test case (10x10 grid, coastal land strip, 3 cardinal boundaries + 1 interior segment), ran `configure_forcings`/`process_forcings` against real GLORYS data, then `case.build`/`case.submit` on Derecho -- completed a full 7-day MOM6 simulation successfully.


manishvenu:
@alperaltuntas This is a first attempt and there seems to be all sorts of issues with interior boundary segments and the regional-mom6 PR has to be merged first. But this is ready for a look and has successfully run a domain.

manishvenu:
Superseded — re-landed as #269 on top of #265/#266.

PR #257, opened 2026-07-23
Divorce OBC/IC engines from MOM6; wire mom6/cice/ww3 through shared GET→REGRID→MERGE pipeline

manishvenu:
## Summary

Divorces MOM6 out of the shared forcing engines so CICE and WW3 can reuse them too, adds fast synthetic "reference" products for exercising all three engines with zero network/credential/`/glade` dependency, and fixes a handful of bugs only a real end-to-end CESM run surfaced.

### Engine split
- **`obc.py`/`ic.py` are now model-agnostic**: GET (download) + MERGE only; REGRID is a caller-supplied callback (`regrid_chunk_fn`/`regrid_fn`) over a plain in-memory `xr.Dataset`.
- **New `mom6.py`** holds everything MOM6-specific (download-request builder, both regrid callbacks, `process_mom6_obc`/`process_mom6_ic`) behind the same external signatures `driver.py` already called.
- **`cice.py` now generates a real forcing file**: a single expanded-grid restoring file (real global CICE restart, windowed to the case domain + a configurable halo via new `mom6_forge.SupergridBase.expand()`, regridded via ESMF `nearest_s2d`). Registry-driven (`cice_product_name`/`cice_function_name`/`cice_function_args`), like MOM6/WW3 already were.
- **`ww3.py`** routes through the shared engine, sourcing real ERA5 2D wave spectra by default (new `raw_data_access/datasets/era5.py`, decoded via `eccodes`). Every real ERA5 grid point in a boundary's window becomes its own station; `ww3_bounc` linearly interpolates between them. The earlier zero-credential placeholder product is removed — WW3 cases need a real CDS key, or use the new `reference_waves` product below.
- **Output split into `ocean/`/`sea_ice/`/`wave/`** (was one shared `ocnice/` dir — WW3's multi-station output alone can be hundreds of files).
- **`base.py`**: `ForcingProduct` backed out to a generic bounding-box-download contract; MOM6-specific var-name fields moved to new `MOM6ForcingProduct`, with `CICEForcingProduct`/`WW3ForcingProduct` extension points alongside it.

### Fast synthetic "reference" products
- **New `raw_data_access/datasets/reference.py`**: `reference_ocean`, `reference_ice`, `reference_waves` — pure numpy/xarray, deterministic (same bbox+dates → identical output), no network/`/glade` dependency. Coarse by default (cheap regrid), overridable resolution.
- New tests: per-product registration/shape/determinism checks, per-engine real end-to-end pipeline tests, and `tests/test_full_workflow.py` (all three engines together via reference products).
- Deleted `test_obc_regrid_workflow` (dead-wired; superseded).

### Fixes found by actually running a case (not just processing forcings)
- **MOM6/WW3 output files were stamping a calendar name FMS rejects at runtime** (`"standard"`, CF-valid but not accepted by MOM6's `get_cal_time`) — every case was crashing at OBC init; only surfaced once a full model run was attempted. `Calendar` gains a `mom6` field (`"gregorian"`/`"noleap"`) separate from the CF-correct `cf` field.

### Companion PRs
- PR in `ESMCI/visualCaseGen#45`: `get_custom_ocn_grid_path()` and two other call sites still pointed at the old combined `ocnice/` dir, which no longer exists — silently broke `case.setup` for every custom-grid case (swallowed by a broad `except Exception`, so it never raised). Also bumps visualCaseGen's embedded `mom6_forge` to pick up `SupergridBase.expand()`.
- - PR in `NCAR/mom6_forge#122`: New `mom6_forge` function `SupergridBase.expand()` to do the CICE grid expansion


manishvenu:
Superseded — re-landed as a stack of smaller PRs on top of #263: #265 (engine split), #266 (forcing_configurations+extract_forcings merge), #267 (CICE), #268 (WW3).

PR #258, opened 2026-07-23
Update CICE settings

dabail10:
I have added the CICE settings to use open boundaries. This relies on the following CESM_CICE tag:

https://github.com/ESCOMP/CESM_CICE/commits/cesm3_cice6_6_3_18/


manishvenu (approved):
Looks good to me! If you don't mind, I can update the branch and merge it.

dabail10:
The Sphinx documentation failed. If you are happy to make a new CESM tag, that would be great. However, should we not coordinate with CESM here? Or are do you have a separate branch?

manishvenu:
@dabail10 We can keep it in PR until the CESM is updated!

PR #262, opened 2026-08-07
Speed up CI: micromamba env caching, drop container from pytest/docs

manishvenu:
**Descrip:**

Speeds up CI by caching the conda environment and dropping the crocontainer image from the two jobs that never needed it.

**Changes:**

1. `raw_data_access_testing.yml`, `deploy_sphinx_docs.yml`, `testing.yml`: switch from `conda-incubator/setup-miniconda` (or a hand-rolled conditional `conda env create`) to `mamba-org/setup-micromamba` with `cache-environment` and `cache-downloads`, keyed on `environment.yml`'s contents.

2. `deploy_sphinx_docs.yml`: drop the crocontainer image entirely. Sphinx autodoc imports CrocoDash but never touches CESM/CIME, so it only ever needed the conda env.

3. `testing.yml` (pytest job): drop the crocontainer image too. `generate_small_test_case.py` only calls CIME's `create_newcase` (pure Python/XML) and the regridding pipeline, never `case.build`/`case.submit` — so it needs a CESM source checkout but not the container's compiler/MPI toolchain. Adds a pinned `CESM_REPO`/`CESM_REF` checkout, cached via `actions/cache` as the full post-`git-fleximod` tree (that update step, not the initial clone, is the expensive part).

**Minor changes:**

- `testing.yml`'s `run_cesm_crocontainer` job is untouched — it does a real `case.build`/`case.submit` via `run_case.sh` inside the container/Apptainer sandbox, which does need the real compiler/MPI stack.


.github/workflows/testing.yml
```diff
@@ -5,99 +5,120 @@ on: [push, workflow_dispatch]
...
-      - name: Create CrocoDash environment if changed
-        if: steps.compare.outputs.env_changed == 'true'
+
+      # The clone itself is cheap -- the expensive part is `git-fleximod
+      # update`, which pulls in every CESM external (MOM6, CICE, CDEPS,
+      # cime, ccs_config, ...). Cache the whole post-fleximod CESMROOT tree
+      # so a cache hit skips both steps entirely.
+      - name: Cache CESM checkout + externals
+        id: cesm-cache
+        uses: actions/cache@v4
```
alperaltuntas, line 53: If `git-fleximod update` below fails partway through, say on a transient network error, the post-job save step still runs (cache skips the save only on cancellation, not on failure), so a half-populated `CESMROOT` gets written under this key. Every run after that hits the cache, skips the checkout step, and then fails later with a confusing missing-externals error that nobody can clear without bumping the key. Can we split this into `actions/cache/restore` plus an explicit `actions/cache/save` with `if: success()`, or drop a sentinel file after fleximod succeeds and re-clone when it isn't there?

.github/workflows/testing.yml
```diff
@@ -5,99 +5,120 @@ on: [push, workflow_dispatch]
...
+        with:
+          path: ${{ env.CESMROOT }}
+          key: cesm-${{ runner.os }}-${{ env.CESM_REF }}-v4
```
alperaltuntas: `CESM_REF` is a branch, not a tag or a SHA, but the key is fixed, so after the first save every run gets a hit and skips the clone entirely. New commits on `full_regional_cesm` would then never make it into CI until someone remembers to bump the `-v4` suffix by hand. Can we resolve the SHA with a `git ls-remote` step and fold that into the key instead (or pin `CESM_REF` to a tag)?

manishvenu: Done with finding the sha

.github/workflows/testing.yml
manishvenu: Done with if success

alperaltuntas (approved):

Issue #261, opened 2026-08-07
Weird Behavior and Suggestions for CrocoDash CLI and Large Data Workflow (LDW)


AidanJanney:
In the new LDW that is orchestrated through the CrocoDash cli, I have a few issues/suggestions that I wanted to collect here:

**Issues**
- `start_date` and `end_date` appear twice in `extract_forcings/config.json`. It is not immediately clear which to change and why it shows up twice. 
- The cf_calendar option is set to "standard" for cf_time. This does not correctly correspond to the GREGORIAN calendar in CESM (which is actually proleptic gregorian - an outstanding issue for CESM/MOM which need to change their label to 'proleptic_gregorian' to match CF conventions, https://earthsystemmodeling.org/docs/nightly/develop/ESMF_usrdoc.pdf). The mismatch is likely neglegible for runs after the 16th century, but worth noting. However, "standard" calendar is not recognized by CESM/ESMF, so this throws a FATAL error. We should change this to "gregorian" so CESM can recognize it. This is also the value that GLORYS uses.
- The `step` value in `extract_forcings/config.json` does not affect the retrieval step for get_glorys_data_from_rda (and possibly other sources). It does apply on the regrid step but is missing on the retrieval step. 

**Suggestions**

- The new format with the CLI works great, but we should provide a short readme with guidance in the extract_forcings folder. 
- The change to split settings between config.json and crocodash_state.json makes LWD a bit more confusing. Perhaps we could symlink or at least explicitly point to crocodash_state.json so users know what file needs to be changed.

manishvenu:
> The cf_calendar option is set to "standard" for cf_time. This does not correctly correspond to the GREGORIAN calendar in CESM (which is actually proleptic gregorian - an outstanding issue for CESM/MOM which need to change their label to 'proleptic_gregorian' to match CF conventions, https://earthsystemmodeling.org/docs/nightly/develop/ESMF_usrdoc.pdf). The mismatch is likely neglegible for runs after the 16th century, but worth noting. However, "standard" calendar is not recognized by CESM/ESMF, so this throws a FATAL error. We should change this to "gregorian" so CESM can recognize it. This is also the value that GLORYS uses.


I have a fix for this if you want it!

> start_date and end_date appear twice in extract_forcings/config.json. It is not immediately clear which to change and why it shows up twice.

Changing the conditions appears to have made this show up twice.

[...]

PR #263, opened 2026-08-10
Fix MOM6/CESM calendar-name FMS rejection; config.json cleanups (#261)

manishvenu:
**Descrip:**

Fixes MOM6/CESM output files stamping a calendar name FMS rejects, plus two config bugs from #261. Extracted from #256 and #257, which both independently made the same calendar fix — pulled out here so both can stack on it instead of duplicating it.

**Changes:**

1. Calendar name `"standard"` → `"gregorian"`/`"noleap"`. Closes the calendar half of #261.

2. Dedupe `start_date`/`end_date`, previously duplicated under both `inputs` and `outputs` in `extract_forcings/config.json`, and split the single `step` value into `get_step_days`/`regrid_step_days` so the GET phase actually chunks instead of silently ignoring it. Addresses the other two bugs in #261.

**Minor changes:**

- Rename `crocodash_state.json` → `_crocodash_state.json` to signal it isn't meant to be hand-edited. Addresses the suggestion in #261.

**Calendar handling (rework following review):**

- `ForcingProduct` now requires `calendar` to be a `Calendar` instance. The `cf`/`cesm`/`mom6` names can no longer be declared individually, so they cannot drift apart — a product declares one constant (`GREGORIAN`, `NOLEAP`) and the three names follow.
- `write_metadata` expands dataclasses into named dicts, so the `Calendar` survives into the product metadata block of `config.json` as a nested object instead of being dropped by that method's JSON filter; the OBC step reads `calendar.mom6` straight out of it. The chl and river-nutrients configurators store their own copy in their `inputs` block, and that is what `Calendar.from_config` reads back in the extraction driver.
- Writers take the `Calendar`, not a bare string. `bgc.process_river_nutrients` uses `.cf` for its `cftime` calls and `.mom6` for the attribute it writes to disk — two roles one string could not serve. `process_chl` and `convert_cftime_to_numeric` likewise name the role they need, replacing hardcoded defaults.
- `Case.configure_forcings` now `xmlchange`s `CALENDAR`. It previously had no consumer anywhere, so CESM stayed on its `NO_LEAP` default while GLORYS forcing was stamped Gregorian — which FATALs in `get_cal_time` on any run crossing Feb 29.

**Open boundary conditions:**

- BGC tracer OBC params are emitted once per tracer rather than once per boundary, since each tracer lives in a single `<tracer>_obc_segment.nc` spanning all segments.
- Added `_split_bgc_tracers_into_files`, which writes those per-tracer files. `regional_mom6`'s equivalent only runs inside its own `setup_ocean_state_boundaries` and is never reached by this pipeline, so without this the files MOM6 is pointed at never exist.
- The split opens each per-boundary file once rather than once per tracer, and streams `to_netcdf` from the open sources.
- The BGC tracers are now *moved*, not copied: they are dropped from `forcing_obc_segment_NNN.nc` after the per-tracer files are written. `OBC_SEGMENT_NNN_DATA` names only U/V/SSH/TEMP/SALT from that file, so the copies were never read and were doubling the largest files on disk. The split skips when the per-tracer files already exist, so resume still works.
- Removed a dead `calendar`/`time_units` kwargs block that would have raised `TypeError` for duplicate keyword arguments had it ever fired.
- `write_metadata` drops `marbl_var_names` when the BGC tracers were not merged, so the key means "these were regridded" rather than merely "this product declares some".

**Other:**

- GLORYS RDA access globs the year subdirectory matching each date rather than walking the whole multi-decade archive, and now raises with the date and pattern when a date has no match instead of silently yielding a dataset with a gap in time.
- `ask_yes_no` returns its `default` on EOF rather than always `False`, so a piped `crocodash fork` is usable non-interactively. The four `_resolve_copy_plan` prompts pass `default=False` so that change does not silently copy SourceMods or apply xmlchanges unconfirmed.

**Breaking changes:**

- Any out-of-repo `ForcingProduct` must declare `calendar = <Calendar>`; declaring `cf_calendar`/`cesm_calendar`/`mom6_calendar` directly now fails at class-definition time.
- `config.json` files written before this carry flat `*_calendar` keys and will not be read by the new `Calendar.from_config`. Re-run `configure_forcings` to regenerate.


CrocoDash/raw_data_access/base.py
```diff
@@ -29,14 +29,23 @@
...
 @dataclass(frozen=True)
 class Calendar:
-    """Pairs a CF time-attribute calendar name with its corresponding CIME CALENDAR xml value, so a product can't declare one without the other."""
+    """Pairs the calendar name each downstream consumer expects, so a product can't declare one without the others.
 
-    cf: str  # CF name for dataset time attrs
-    cesm: str  # CIME CALENDAR xml value
+    cf is for xarray's own time decode/encode while reading raw data (its CF
+    convention name, e.g. "standard" for the real-world calendar). cesm is
+    the CIME CALENDAR xml value. mom6 is the literal string the regridding
```
alperaltuntas: Can we set CALENDAR from cesm_calendar somewhere, maybe next to the RUN_STARTDATE and STOP_OPTION xmlchanges in case.py:538? Adding mom6 as a third field here is the right place for it, but cesm still has no consumer, on this branch or on main. Nothing calls xmlchange on CALENDAR, and it defaults to NO_LEAP in cmeps config_component.xml. So for GLORYS we now stamp gregorian on the OBC files, and get_cal_time also checks the file's calendar against whatever time_manager is using (get_cal_time.F90 line 203) and aborts if they don't match. I think that means GLORYS cases still won't get past OBC init, just with a different FATAL than the standard one. Can you double check please?

Also see cesm_ocean_output.py line 554, where noleap is hardcoded while line 52 of the same file declares calendar = NOLEAP. Same in extract_forcings/bgc.py line 83 and extract_forcings/chlorophyll.py line 10, which both default independently of the product. Can these read from the bundle instead? These are the scattered calendar specifications I was asking about on PR 220.

CrocoDash/extract_forcings/driver.py
alperaltuntas: Can we switch this and line 249 to mom6_calendar? These are two of the sites I listed on PR 220, the chl path and extract_forcings/bgc.py, and they still take cf_calendar, which is standard for GLORYS, so both files hit the same get_cal_time FATAL the OBC files do. Though see my comment on base.py for why I don't think patching the field name at each call site is the fix. For a more general fix, please see my comment regarding the design of the Calendar class.

CrocoDash/raw_data_access/base.py
```diff
@@ -29,14 +29,23 @@
...
+    step must stamp on output forcing files' time:calendar attribute, since
+    MOM6's get_cal_time() accepts neither cf's "standard" nor cesm's
+    upper-cased "GREGORIAN"/"NO_LEAP" -- only e.g. "gregorian"/"noleap".
```
alperaltuntas: Do you mind rewriting this docstring without using Claude? In its current shape, I can't fully follow it. Also, the upper NO_LEAP isn't what breaks it. NO_LEAP fails because the underscore form isn't in the accepted list, and standard fails because it isn't in the list at all.

CrocoDash/shareable.py
```diff
@@ -710,8 +710,13 @@ def ask_yes_no(prompt: str, default=True) -> bool:
...
     except EOFError:
-        print("No input available, assuming 'no'.")
-        return False
+        # Fall back to `default`, matching ask_string() above and matching what
+        # pressing enter at an interactive prompt would do. Returning False
+        # regardless of the default made ForkBundle.fork() unusable
+        # non-interactively: its final "Proceed with this configuration?" is
+        # default=True, so an EOF turned into "Fork cancelled by user."
+        print(f"\nNo input detected, using default: {default!r}")
+        return default
```
alperaltuntas, line 723: Doesn't this flip the four _resolve_copy_plan prompts from no to yes? They call ask_yes_no without a default, so on EOF a piped crocodash fork would copy SourceMods and run apply_xmlchanges_to_case with nobody confirming. Can we pass default=False at those call sites, or check sys.stdin.isatty() here?

CrocoDash/extract_forcings/obc.py
alperaltuntas: This sets kwargs["calendar"] and kwargs["time_units"], and the call at line 336 now passes both explicitly plus **kwargs, so if this branch ever fires we get a TypeError for multiple values. It's dead now because products declare calendar as a Calendar instance and is_json_compatible drops it from the metadata. So can we just remove the block?

CrocoDash/raw_data_access/datasets/glorys.py
```diff
@@ -80,8 +80,12 @@ def get_glorys_data_from_rda(
         date_strings = [date.strftime("%Y%m%d") for date in dates]
 
         for date in date_strings:
-            pattern = os.path.join(ds_in_path, "**", f"*_{date}_*.nc")
-            ds_in_files.extend(glob.glob(pattern, recursive=True))
+            # Files are filed under a year subdirectory matching their own date
+            # (e.g. d010049/1993/..._19930909_...nc). Scoping the glob to that
+            # subdirectory avoids a recursive walk of the entire multi-decade
+            # archive for every single day requested.
+            pattern = os.path.join(ds_in_path, date[:4], f"*_{date}_*.nc")
```
alperaltuntas, line 87: If the glob comes back empty for a date we just carry on, and open_mfdataset then either fails with no files to open or hands back a dataset with a gap in time. Can we raise here with the date and the path when there's no match?

CrocoDash/forcing_configurations/configurations.py
```diff
@@ -711,15 +714,15 @@ def configure(self):
...
-        self.set_output_param("end_date", end_date)
         self.set_output_param(
             "information",
             product.write_metadata(include_marbl_tracers="%MARBL" in compset),
         )
         start_dt = datetime.strptime(start_date, self._DATE_FORMAT)
         end_dt = datetime.strptime(end_date, self._DATE_FORMAT)
-        self.set_output_param("step", (end_dt - start_dt).days + 1)
+        step = (end_dt - start_dt).days + 1
+        self.set_output_param("get_step_days", step)
```
alperaltuntas, line 744: `step` here is the whole date range, `(end_dt - start_dt).days + 1`, and both keys get it. So `_make_date_pairs` is handed a chunk size equal to the entire period, its loop runs once, and `GET` still does a single request covering everything, same as before.

Wiring `get_step_days` through to `_get_boundary` is the part issue #261 was missing, so that half is right. But with the value set this way I don't think anything actually changes. What should `get_step_days` be? Something the user sets through `configure_forcings`, or something we derive from the product's own download limits?

CrocoDash/extract_forcings/obc.py
```diff
@@ -382,6 +382,55 @@ def _merge_boundary(boundary_label: str, regridded_files: list, output_folder) -
...
+    this pipeline.
+    """
+    if not marbl_var_names:
+        return []
+
+    output_path = Path(output_path)
+    seg_ids = [f"{n:03d}" for n in boundary_number_conversion.values()]
+    written = []
+
+    for var in marbl_var_names:
```
alperaltuntas: This reopens every forcing_obc_segment_NNN.nc once per tracer. For CESM_POP_OUTPUT with four boundaries that's 30x4 opens of the biggest files in the pipeline, and the .load() calls hold all four segments of a tracer in memory at once. Can we open the segment datasets once outside the loop and let to_netcdf stream, the way rm6 does?

CrocoDash/extract_forcings/obc.py
```diff
@@ -382,6 +382,55 @@ def _merge_boundary(boundary_label: str, regridded_files: list, output_folder) -
...
+                dz_var_name = f"dz_{var_name}"
+                if dz_var_name in ds:
+                    ds_var[dz_var_name] = ds[dz_var_name].load()
+
+        out_file = output_path / f"{var}_obc_segment.nc"
+        ds_var.to_netcdf(out_file, unlimited_dims="time")
+        written.append(out_file)
+        logger.info("BGC SPLIT: wrote %s", out_file.name)
+
+    return written
```
alperaltuntas, line 470: The tracers get copied out but stay in forcing_obc_segment_NNN.nc, and the OBC_SEGMENT_NNN_DATA string only names U/V/SSH/TEMP/SALT from that file, so every BGC field ends up on disk twice. With 30 tracers and a year of daily data that adds up. We need to drop them from the per-boundary file, I think.

CrocoDash/raw_data_access/base.py
alperaltuntas: merged is computed on every call but only used in the first branch, and base["marbl_var_names"] at line 262 is already set by the cls.__dict__ sweep in BaseProduct.write_metadata.

CrocoDash/raw_data_access/base.py
```diff
@@ -29,14 +29,23 @@
 
 @dataclass(frozen=True)
 class Calendar:
```
alperaltuntas, line 32: The aggregation only gets us half of what I was after on 220. It stops the three names from disagreeing where they're declared, but the failure mode is a call site reaching for the wrong one, and nothing here makes that harder. This is the third time we're fixing an instance of it.

Two things I'd consider:

(1) cesm isn't the same kind of thing as cf and mom6. Those two are strings we stamp into a netCDF time attribute, cesm is an action on the case. Flattening all three into peer fields is part of why nothing ever called xmlchange on it. Making it a method would at least make the omission visible:

    def apply_to_case(self, case):
        xmlchange("CALENDAR", self.cesm, is_non_local=case.cc._is_non_local())

(2) The writers should take the Calendar, not a string. extract_forcings/bgc.py shows why: it takes one calendar argument and uses it for cftime.datetime at line 137, date2num at line 183, and the time attr it writes at line 204. Those are two different roles, cftime's and FMS's, and one string can't be right for both by construction. If bgc.py took the Calendar, each of those lines could name what it needs.

alperaltuntas (changes requested):

CrocoDash/extract_forcings/obc.py
manishvenu: Removed this.

CrocoDash/raw_data_access/datasets/glorys.py
manishvenu: yes, done!

CrocoDash/raw_data_access/base.py
manishvenu: Sure, this makes sense, for some reason the marbl_var_names were not showing up at some point, but they should, so I'm removing it for now.

CrocoDash/shareable.py
manishvenu: Passed default = False

CrocoDash/raw_data_access/base.py
manishvenu: Yes done.

CrocoDash/forcing_configurations/configurations.py
manishvenu: So having GET be a single request is actually fine I think for now. Doing 5-day increments is not necessarily faster, so I'm having the default be a single request for GET. Maybe to keep the complexity down, I'll open a new PR to add a download limits feature on the products

manishvenu: *I can open

CrocoDash/extract_forcings/obc.py
manishvenu: Fixed.

CrocoDash/extract_forcings/obc.py
manishvenu: added a drop loop to the split function.

CrocoDash/extract_forcings/driver.py
manishvenu: Ok yes, I'll pass in the Calendar Class.

CrocoDash/raw_data_access/base.py
manishvenu: I thoguht we did set the calendar, but we did not so I do add it to that case location. I'll get them to read from the bundle instead.

CrocoDash/raw_data_access/base.py
manishvenu: The apply_to_case is a bit confusing on Calendar class, so maybe will pass on this for now?  I did implement 2. 

One thing that came up while doing this is to have the Calendar class be an attribute on forcing products instead of having three seperate calendar attributes.

alperaltuntas (approved):

PR #265, opened 2026-08-11
Split OBC/IC extraction into model-agnostic GET/REGRID/MERGE engines

manishvenu:
**Descrip:**

Splits MOM6 out of `extract_forcings/obc.py` / `ic.py` so CICE and WW3 can reuse the same GET → REGRID → MERGE pipeline.

**Changes:**

1. REGRID becomes a caller-supplied callback (`regrid_chunk_fn`/`regrid_fn`) over a plain in-memory dataset; all MOM6-specific logic moves into a new `mom6.py`.

2. Adds `REFERENCE_OCEAN`, a fast synthetic MOM6 forcing product (no network, no credentials) for tests and demos, *This is completely fake data, can completely be ignored. It just needs to run**
3. splitting `ForcingProduct` into `ForcingProduct` → `VelocityTracerForcingProduct` → `MOM6ForcingProduct`. *

**Minor changes:**

- Bumps the embedded `rm6` submodule pointer for the new `regional_mom6` API: `Segment.regrid_velocity_tracers` dropped `rotational_method`, and `experiment` now takes `mom6_forge` `Grid`/`VGrid` objects directly.


manishvenu:
This is an incredibly difficult merge from here up to #266 . I think we need to make sure #270 gets merged before #266

manishvenu:
@alperaltuntas So do you mind reviewing this PR and #270 before #266? I have to manually go through and make sure we haven't dropped any new code because of the refactoring in #266

CrocoDash/extract_forcings/obc.py
alperaltuntas: This is still `product_info`, but the arg is `dataset_varnames` now. Please run all tests (including slow) to catch any similar issues.

alperaltuntas: Isn't this the same function that now lives in `mom6.py`?

tests/extract_forcings/test_obc.py
alperaltuntas: Same three tests as in `test_mom6.py`. Please remove this and any other duplications (in the tests or elsewhere).

CrocoDash/extract_forcings/mom6.py
```diff
@@ -0,0 +1,516 @@
...
+
+    output_path = Path(output_path)
+    seg_ids = [f"{n:03d}" for n in boundary_number_conversion.values()]
+    seg_files = {seg: output_path / f"forcing_obc_segment_{seg}.nc" for seg in seg_ids}
+    out_files = [output_path / f"{var}_obc_segment.nc" for var in marbl_var_names]
+
+    # Step 3 strips the BGC tracers out of the per-boundary files, so on a
+    # re-run they are no longer there to be read. Every other phase of the OBC
+    # pipeline resumes by skipping completed work; do the same here.
+    if all(f.exists() for f in out_files):
```
alperaltuntas, line 158: If this gets killed partway through the `to_netcdf` loop, the last file exists but is truncated. Then on a re-run we skip the split here, skip the merge too (output is already there), and hand MOM6 a corrupt file without any complaint. The other resume checks call `utils.is_valid_netcdf()` and raise. Can we do that here too?

alperaltuntas (changes requested):
In addition to the line comments below, one thing on backward compatibility: this PR renames `crocodash_state.json` to `_crocodash_state.json`, moves `start_date`/`end_date` from `conditions.outputs` to `conditions.inputs`, and replaces `step` with `get_step_days`/`regrid_step_days`. But `SCHEMA_VERSION` in `case_state.py` is still 1.0.0. Can we bump the version?

CrocoDash/extract_forcings/obc.py
manishvenu: I think this is still marbl_var_names according to raw data. I've removed this function because it overlapped with MOm6.

manishvenu: Yes, removed

manishvenu: It was a leftover from the merge I think.

tests/extract_forcings/test_obc.py
manishvenu: Done, I believe there aren't anymore/

manishvenu:
I think there are some pre-existing failures but thetests are passing for this section (pytest doesn't work right now because the container has a different cime version)

CrocoDash/extract_forcings/mom6.py
manishvenu: Done

alperaltuntas (approved):

manishvenu:
Thanks!

PR #266, opened 2026-08-11
Merge forcing_configurations + extract_forcings into forcing/

manishvenu:
**Descrip:**

Merges `forcing_configurations/` (declarative config classes) and `extract_forcings/` (process functions) into one `forcing/` package. Each forcing type is now a single class owning both its configuration and its extraction logic, declared via a `process_components = {flag: method}` manifest.

**Changes:**

1. `driver.py` / `cli.py` / `Case.process_forcings()` dispatch generically off that manifest instead of a hand-maintained `if`-ladder per forcing type. Adding a new forcing type is now a one-file (+ tests) change.

2. `forcing/__init__.py` auto-discovers configurator modules via `pkgutil`, matching the existing pattern in `raw_data_access/datasets`.

3. The `BGCRiverNutrients` → `Runoff` cross-configurator dependency is now declared via `depends_on_outputs` and validated by a test, replacing an undeclared raw reach into `ctx.config["runoff"]["outputs"][...]`. `driver.py` derives its run-before ordering from that declaration instead of a separately hand-maintained list.

**Minor changes:**

- Drops the grouped `process_bgc` flag in favour of independent `process_bgcic` / `process_bgcironforcing` flags.
- `BaseConfigurator` exposes an `is_non_local` property sourced from the registry's live `Case`, applied automatically to any `XMLConfigParam` output — a configurator no longer threads its own `case_is_non_local` ctor arg just to reach `apply()` with the right `--non-local` behaviour.
- `get_input_param` / `get_output_param` raise a dedicated `UndeclaredParamError`, so a stale or renamed param reference can't be mistaken for an incidental `KeyError` from a configurator's own logic.

---

**⚠️ Reviewers / anyone merging into this branch: git cannot follow this diff.**

`forcing_configurations/` + `extract_forcings/` were merged into `forcing/` as a rename *and* a rewrite (process functions became methods on their configurator class), so git's rename detection gives up and reports the old files as deleted and the new ones as added. Two consequences:

1. **The raw diff is not reviewable as a diff.** Reviewing "what actually changed" means comparing function-by-function, not reading `git diff`. For reference, an AST-level comparison of the two trees came out at 94 functions byte-identical after dedent, with every `input_params` / `output_params` declaration preserved verbatim — the real changes are the ones listed above.

2. **Merges from the base branch silently drop changes to the old paths.** Anything landing on `refactor-obc-ic` that touches `extract_forcings/` or `forcing_configurations/` merges cleanly here *as a no-op*, because those files no longer exist on this branch — no conflict, no warning. Such changes have to be re-applied to `forcing/` by hand.


manishvenu:
This cannot be merged before #270 #265 are merged.

manishvenu:
@alperaltuntas 

Claude's listed actual changes:

- driver.py’s new dispatch loop
- base.py’s five new registry methods (to handle the processing)
- The process_forcings kwarg change in case.py. 

That is under 300 lines, and it is the entire coding content of the pull request.

CrocoDash/case.py
```diff
@@ -595,24 +592,14 @@ def process_forcings(
...
-            runoff=process_runoff and self.fcr.is_active("runoff"),
-            bgcrivernutrients=process_bgc_river_nutrients
-            and self.fcr.is_active("BGCRiverNutrients"),
-        )
+        config_path = self.extract_forcings_path / "config.json"
+        with open(config_path) as f:
+            config = json.load(f)
+
+        flag_names = ForcingConfigRegistry.resolve_process_targets(config).keys()
+        flags = {name: kwargs.get(f"process_{name}", True) for name in flag_names}
```
alperaltuntas, line 614: `process_forcings` only takes `**kwargs` now, and nothing checks what's left over. So if
a script passes `process_initial_condition=False`, which was a real argument before this
PR, it just goes into `**kwargs`, nothing reads it, and `kwargs.get("process_ic", True)`
returns True, so the IC gets regenerated instead of skipped. Same for
`process_velocity_tracers`, `process_bgc` and `process_bgc_river_nutrients`, and for a
typo like `process_ics`.

Nothing in the repo or in the gallery notebooks passes these, so CI stays green. Can we
raise an error on `process_*` kwargs we don't recognize, so these fail instead of
quietly doing the opposite?

CrocoDash/forcing/driver.py
```diff
@@ -0,0 +1,160 @@
...
+                    f"[info] '{flag_name}' requires '{dep}' to run first -- "
+                    "enabling it automatically"
+                )
+                requested.add(dep)
+
+    if not requested:
+        print("No components selected.")
+        return
+
+    # Stable order: anything named as a dependency runs before its dependents.
```
alperaltuntas: This only puts a flag's direct deps first, and it loops over `requested`, which is a
set, so the order isn't actually stable. I tried a three level chain (chaina needs
chainb, chainb needs chainc) and got `['chainb', 'chaina', 'chainc']` with
PYTHONHASHSEED 0 and 1, and the right order `['chainc', 'chainb', 'chaina']` with 2, 3
and 4. The auto-enable loop above only goes one level deep too: if config.json lists
chainb before chaina, asking for chaina turns on chainb but never chainc.

We only have one edge right now (bgcrivernutrients needs runoff) so nothing is broken
yet. `order_overrides` is already the `{node: predecessors}` dict that
`graphlib.TopologicalSorter` takes, so can we just pass it there and use
`static_order()`? That sorts the whole chain, and raises CycleError, which this loop
can't catch.

alperaltuntas (changes requested):

CrocoDash/case.py
manishvenu: Done

CrocoDash/forcing/driver.py
manishvenu: Done

alperaltuntas (approved):

PR #268, opened 2026-08-11
Add WW3 forcing to the merged forcing/ pattern

manishvenu:
**Descrip:**

Adds WW3 boundary-spectra forcing on the merged `forcing/` pattern from #265/#266: a `WW3Configurator` (`forcing/ww3.py`) routing through the shared GET/REGRID/MERGE engine via its own regrid callback, same as MOM6's OBC.

**Changes:**

1. Adds `WW3ForcingProduct`, `era5.py` (real ERA5 2D wave spectra via CDS), and `REFERENCE_WAVES` (synthetic, no-network stand-in for tests and demos).

**Minor changes:**

- No changes needed to `driver.py` / `cli.py` / `case.py`.


alperaltuntas:
Pushed two commits to `ww3-new` and retargeted the PR to `main` now that `class-merge` is in:

- `2838dc54` merges `origin/main` in. The 8 conflicting files that this PR does not touch (case.py, bgc.py, driver.py, the tests, the glade QC script) were taken from `main`; `CrocoDash.forcing.rst` keeps both the new `atm`/`cdeps_streams` entries from `main` and the `ww3` entry from here.
- `39437123` changes `WAVE_SUBDIR` from `"wave"` to `"wav"`. `case.py` writes the WW3 grid inputs to `inputdir/wav` and `WW3_GRID_INP_DIR` points there, so `ww3_bounc.nml` written to `wave/` was never picked up by the WW3 buildnml. The three tests that hardcoded `tmp_path / "wave"` now use `WAVE_SUBDIR`.

`test_ww3 test_era5 test_reference test_configurators test_cli`: 53 passed, 1 skipped. The PR diff against `main` is unchanged otherwise (13 files).

Related: the WW3 grid needs status-2 boundary points for nest.ww3 to be used at all; that landed in NCAR/mom6_forge#139/#140 and is pinned into CrocoDash via #282.

alperaltuntas:
Two more commits on `ww3-new`, from a 31-day plumbing test of this PR (`reference_waves` on all four boundaries of the agulhas6 grid, with the status-2 boundary ring from #282):

- `adaaba21` merges `main` again to pick up #282.
- `d066d511` fixes a crash. `REFERENCE_WAVES` spread its three stations along longitude at one mid-latitude, so on the west/east boundary windows all three landed on the same point. `ww3_bounc` interpolates each boundary point from its two nearest stations with distance weights, and for coincident stations those are NaN: 72 of 248 boundary points had `RDBP = NaN` in `nest.ww3`, the boundary spectra went NaN, and the model segfaulted in `w3sin4` on step 1. The stations are now spread along the long axis of the window, and `process()` writes each station location once (adjacent boundaries share their corner stations). The `multi_station` test's synthetic product gets the same change.

With the fix the run completes with boundary input active throughout (124 six-hourly updates, `FLBPI` never dropped), the fetch ramp at the rim is gone, and long-period energy enters at the boundary as expected. 33 tests pass.

Two things to know when regenerating spectra on an existing case: `process --ww3` skips the GET/REGRID/MERGE when `extract_forcings/ww3/{raw_data,regridded_data,merged}` exist, and the WW3 buildnml copies `spec.list`/`ww3.point*_spec.nc` into `<rundir>/ww3_moddef_create/` only when missing, so stale copies make `ww3_bounc` fail with `SPEC FILE DOES NOT EXIST`.

alperaltuntas:
`464f2bc3` addresses the review comments:

- **eccodes** declared in `environment.yml` (`python-eccodes`); it was imported by `era5.py` but never installed.
- **HIST_OPTION/HIST_N** no longer set by the configurator. They are case-wide and forced hourly `cpl.hi` on every WW3 case. Caveat: the WW3 cap ignores its own namelist `history_option` and only honours these variables, so a case that wants wave history has to set them itself (`./xmlchange HIST_OPTION=nhours,HIST_N=3`).
- **Longitudes**: the CDS `area` is built in -180..180 (a window crossing 180E raises rather than sending W>E), and decoded station longitudes (0..360 from the GRIB) are wrapped onto the WW3 grid's branch before `ww3_bounc` computes distance weights.
- **Land stations** (all-zero over the window) are dropped; with `BOUND%INTERP=2` they damp the adjacent boundary points.
- **`ww3_obc_function_overrides`** added so accessmethod arguments like `cdsapi_rc_path` can be set; `extra_args` was hardcoded to `{}`.
- **Bundle/fork** keep `wav/` next to `ocn/` instead of flattening WW3 inputs into `ocn/`, where `WW3_GRID_INP_DIR` does not look.
- Stale `ww3.point*_spec.nc` removed before writing; `boundaries` type-checked; two docstrings fixed; `xarray` imported in `test_era5.py`.

WW3/ERA5 suites 37 passed, shareable suite 77 passed.

PR #269, opened 2026-08-11
Add interior/custom OBC segment support

manishvenu:
**Descrip:**

Adds interior/custom OBC segment support on the merged `forcing/` pattern from #265/#266, replacing #256. A boundary can now be a live `regional_mom6.segment.Segment` (built via `Segment.from_hgrid`/`from_lonlat`), not just a cardinal string.

**Changes:**

1. Bumps `rm6` to the current tip of the standalone `regional-mom6` repo's `Segment` class (`ocean_side` replacing `mom6_index_reverse`, land-capped-endpoint checks), and migrates OBC regrid and tidal regrid off the old `rm6.segment(...)` factory / `experiment.setup_boundary_tides` onto `Segment.cardinal` / `from_spec` / `regrid_tides` directly.

2. `Case.configure_forcings(boundaries=None)` now auto-detects open cardinal edges from `ocn_topo` instead of hardcoding all four.

**Minor changes:**

- Removes the unused `Case.expt` property — cardinal-only, doesn't generalize to custom segments.

PR #270, opened 2026-08-14
Add parallel fetching of glorys data

enrico-mi:
I am working on adding support to fetch data in parallel for GLORYS so that downloading multiple years is more user friendly.

#### Solution implemented

Multi-processing. `_get_one_chunk` gets one chunk of data and it's isolated to be called by multiple processes at once. glorys.py has minor changes, basically `parallel=False` in `xr.open_mfdataset()` otherwise it silently fails for some reason.

Note that xr.open_mfdataset() calls dask, so there's probably multi-threading also going on within each process.

Number of workers capped at 12 for now but we should evaluate.

I'm testing it over the weekend, but for now it's working (and scaling as much as possible!): 
- single process took 19 mins for one 30-day chunk
- multi process with 6 cpus took 19 mins for 6 30-day chunks

#### Solutions discarded

I tried multi-threading but each thread was competing (for the GIL I think) and mostly waiting, so there was no clear speed gain even if fetching was happening in parallel

#### Next steps

- [x] I'll run it across 12 process over the weekend. Fingers crossed
- [x] Need to test other functions besides `get_glorys_data_from_rda`
- [x] Need to define max workers
- [x] Need to see if regridding and the rest of the processing forcings works fine
- [x] Need not to break pre-existing tests :)


enrico-mi:
post weekend update:

- no code failures
- 12 processors downloaded 207 GB in 25h 38' -> 2.24 MB/s (vs ~0.19 MB/s single proc, 11.8x faster)
- job timed out after 24h before completing all downloads, restart was no issue

manishvenu:
> post weekend update:
> 
> * no code failures
> * 12 processors downloaded 207 GB in 25h 38' -> 2.24 MB/s (vs ~0.19 MB/s single proc, 11.8x faster)
> * job timed out after 24h before completing all downloads, restart was no issue

That's awesome Enrico! That is going to be so cool. If you edit the base branch to calendar-fix it will be easier to see the diff (you can edit the base branch if you click edit on the title) when you are ready to review.

enrico-mi:
> > post weekend update:
> > 
> > * no code failures
> > * 12 processors downloaded 207 GB in 25h 38' -> 2.24 MB/s (vs ~0.19 MB/s single proc, 11.8x faster)
> > * job timed out after 24h before completing all downloads, restart was no issue
> 
> That's awesome Enrico! That is going to be so cool. If you edit the base branch to calendar-fix it will be easier to see the diff (you can edit the base branch if you click edit on the title) when you are ready to review.

Thanks, I wasn't sure whether to target that or main.

I also parallelized `_regrid_boundary()`. I'm testing it at the moment

enrico-mi:
Regridding in parallel also works. For this, the list of files is split in N parts with N number of workers, and each worker works through its list. This way each worker generates the regridding weights for the first chunk (of the side it's working on) and reuses it for the remaining files in its list.

Merging also was succesful.

I'll look at the CI failures next.

enrico-mi:
Hi @manishvenu , I think the functionality is in a good state so now I need pick your brain on a few things:

1. I parallelized only `get_glorys_data_from_rda()`. What do we want to do in the other cases?
     - `get_glorys_data_from_cds_api()`: parallelizing this would be similar to `from_rda()`, so multiple processes would call copernicusmarine package with different start and end dates
     - `get_glorys_data_script_for_cli()`: here we could write a script with multiple lines such that it can be run in parallel from CLI; basically `get_glorys_data()` in `regional_mom6` would append multiple lines, all ending with `&` so that execution will be in parallel if one requests the correct resources

2. no matter what we decide above, `_regrid_boundary()` that is now in parallel should be compatible with all the `get_glorys_data()` (so this gain is independent of the getter!), but I'm a bit unsure about the CLI getter as I don't remember how the regridding part gets the list of files. What do you think?

3. how can I update the documentation (e.g. [this page](https://crocodile-cesm.github.io/CrocoGallery/latest/crocodash/process-forcings/))?

manishvenu:
> Hi @manishvenu , I think the functionality is in a good state so now I need pick your brain on a few things:
> 
> 
> 
> 1. I parallelized only `get_glorys_data_from_rda()`. What do we want to do in the other cases?
> 
>      - `get_glorys_data_from_cds_api()`: parallelizing this would be similar to `from_rda()`, so multiple processes would call copernicusmarine package with different start and end dates
> 
>      - `get_glorys_data_script_for_cli()`: here we could write a script with multiple lines such that it can be run in parallel from CLI; basically `get_glorys_data()` in `regional_mom6` would append multiple lines, all ending with `&` so that execution will be in parallel if one requests the correct resources
> 
> 
> 
> 2. no matter what we decide above, `_regrid_boundary()` that is now in parallel should be compatible with all the `get_glorys_data()` (so this gain is independent of the getter!), but I'm a bit unsure about the CLI getter as I don't remember how the regridding part gets the list of files. What do you think?
> 
> 
> 
> 3. how can I update the documentation (e.g. [this page](https://crocodile-cesm.github.io/CrocoGallery/latest/crocodash/process-forcings/))?

1. That solution for the other glory's works for me! We can modify the CrocoDash part of the cli getter instead of changing regional-mom6 for now. It also needs to work for the CESM ocean options as well which should work by default potentially. I'm a bit worried there will be issues with the glory's from python one but we will see!
2. OBC.py parses date strings in raw data file data names to get the files to be regridded. Does that make sense? I can also dive into this change later today to provide a better answer.
3. In the docs folder! In source, there is a file that corresponds. The docs are rebuilt on commit at the link the bot sends in the PR or you can build yourself with the sphinx commands, make html in the docs folder.


Also happy to hop on a call in the afternoon to chat this out

Thanks a bunch this is huge!

enrico-mi:
>    * `get_glorys_data_from_cds_api()`: parallelizing this would be similar to `from_rda()`, so multiple processes would call copernicusmarine package with different start and end dates=

update: indeed this works out of the box with the parallel implementation I developed for the rda getter; it just passes the different function.

However Copernicus marine throws a 429 error (too many requests) if too many processors are used, and the job is terminated. I think it's too much of a hassle right now to implement some solution with semaphors and retries, so personally I'd leave it as is, and the user will have to be careful (it does not depend on us in the end).

I'm going to work on the CLI script approach now

enrico-mi:
>    * `get_glorys_data_script_for_cli()`: here we could write a script with multiple lines such that it can be run in parallel from CLI; basically `get_glorys_data()` in `regional_mom6` would append multiple lines, all ending with `&` so that execution will be in parallel if one requests the correct resources

this also works now. note that the implementation depends on https://github.com/CROCODILE-CESM/regional-mom6/pull/72

eg 1 month, 10-day steps, `get_glory_data.sh` now looks like this (note `&` at end of lines and `wait` at end of file):
```
#!/bin/bash

copernicusmarine subset --dataset-id cmems_mod_glo_phy_my_0.083deg_P1D-m --variable so --variable thetao --variable uo --variable vo --variable zos --start-datetime 2023-01-01T00:00:00 --end-datetime 2023-01-10T23:59:59 --minimum-longitude -98.98996846773888 --maximum-longitude -37.208560213479366 --minimum-latitude 52.537190110517756 --maximum-latitude 58.61512088772785 --minimum-depth 0 --maximum-depth 6000 -o /glade/work/emilanese/croc_input/dev_GLORY_PARALLEL_CESM-MOM6_NWA12_2023_cli/extract_forcings/raw_data -f north_unprocessed.2023-01-01_2023-01-10.nc &


copernicusmarine subset --dataset-id cmems_mod_glo_phy_my_0.083deg_P1D-m --variable so --variable thetao --variable uo --variable vo --variable zos --start-datetime 2023-01-11T00:00:00 --end-datetime 2023-01-20T23:59:59 --minimum-longitude -98.98996846773888 --maximum-longitude -37.208560213479366 --minimum-latitude 52.537190110517756 --maximum-latitude 58.61512088772785 --minimum-depth 0 --maximum-depth 6000 -o /glade/work/emilanese/croc_input/dev_GLORY_PARALLEL_CESM-MOM6_NWA12_2023_cli/extract_forcings/raw_data -f north_unprocessed.2023-01-11_2023-01-20.nc &


copernicusmarine subset --dataset-id cmems_mod_glo_phy_my_0.083deg_P1D-m --variable so --variable thetao --variable uo --variable vo --variable zos --start-datetime 2023-01-21T00:00:00 --end-datetime 2023-01-30T23:59:59 --minimum-longitude -98.98996846773888 --maximum-longitude -37.208560213479366 --minimum-latitude 52.537190110517756 --maximum-latitude 58.61512088772785 --minimum-depth 0 --maximum-depth 6000 -o /glade/work/emilanese/croc_input/dev_GLORY_PARALLEL_CESM-MOM6_NWA12_2023_cli/extract_forcings/raw_data -f north_unprocessed.2023-01-21_2023-01-30.nc &


copernicusmarine subset --dataset-id cmems_mod_glo_phy_my_0.083deg_P1D-m --variable so --variable thetao --variable uo --variable vo --variable zos --start-datetime 2023-01-31T00:00:00 --end-datetime 2023-02-01T23:59:59 --minimum-longitude -98.98996846773888 --maximum-longitude -37.208560213479366 --minimum-latitude 52.537190110517756 --maximum-latitude 58.61512088772785 --minimum-depth 0 --maximum-depth 6000 -o /glade/work/emilanese/croc_input/dev_GLORY_PARALLEL_CESM-MOM6_NWA12_2023_cli/extract_forcings/raw_data -f north_unprocessed.2023-01-31_2023-02-01.nc &


copernicusmarine subset --dataset-id cmems_mod_glo_phy_my_0.083deg_P1D-m --variable so --variable thetao --variable uo --variable vo --variable zos --start-datetime 2023-01-01T00:00:00 --end-datetime 2023-01-10T23:59:59 --minimum-longitude -98.24 --maximum-longitude -36.09333333333348 --minimum-latitude 4.961175132990646 --maximum-latitude 5.441175132990646 --minimum-depth 0 --maximum-depth 6000 -o /glade/work/emilanese/croc_input/dev_GLORY_PARALLEL_CESM-MOM6_NWA12_2023_cli/extract_forcings/raw_data -f south_unprocessed.2023-01-01_2023-01-10.nc &


copernicusmarine subset --dataset-id cmems_mod_glo_phy_my_0.083deg_P1D-m --variable so --variable thetao --variable uo --variable vo --variable zos --start-datetime 2023-01-11T00:00:00 --end-datetime 2023-01-20T23:59:59 --minimum-longitude -98.24 --maximum-longitude -36.09333333333348 --minimum-latitude 4.961175132990646 --maximum-latitude 5.441175132990646 --minimum-depth 0 --maximum-depth 6000 -o /glade/work/emilanese/croc_input/dev_GLORY_PARALLEL_CESM-MOM6_NWA12_2023_cli/extract_forcings/raw_data -f south_unprocessed.2023-01-11_2023-01-20.nc &


copernicusmarine subset --dataset-id cmems_mod_glo_phy_my_0.083deg_P1D-m --variable so --variable thetao --variable uo --variable vo --variable zos --start-datetime 2023-01-21T00:00:00 --end-datetime 2023-01-30T23:59:59 --minimum-longitude -98.24 --maximum-longitude -36.09333333333348 --minimum-latitude 4.961175132990646 --maximum-latitude 5.441175132990646 --minimum-depth 0 --maximum-depth 6000 -o /glade/work/emilanese/croc_input/dev_GLORY_PARALLEL_CESM-MOM6_NWA12_2023_cli/extract_forcings/raw_data -f south_unprocessed.2023-01-21_2023-01-30.nc &


copernicusmarine subset --dataset-id cmems_mod_glo_phy_my_0.083deg_P1D-m --variable so --variable thetao --variable uo --variable vo --variable zos --start-datetime 2023-01-31T00:00:00 --end-datetime 2023-02-01T23:59:59 --minimum-longitude -98.24 --maximum-longitude -36.09333333333348 --minimum-latitude 4.961175132990646 --maximum-latitude 5.441175132990646 --minimum-depth 0 --maximum-depth 6000 -o /glade/work/emilanese/croc_input/dev_GLORY_PARALLEL_CESM-MOM6_NWA12_2023_cli/extract_forcings/raw_data -f south_unprocessed.2023-01-31_2023-02-01.nc &


copernicusmarine subset --dataset-id cmems_mod_glo_phy_my_0.083deg_P1D-m --variable so --variable thetao --variable uo --variable vo --variable zos --start-datetime 2023-01-01T00:00:00 --end-datetime 2023-01-10T23:59:59 --minimum-longitude -37.68856021347937 --maximum-longitude -36.093333333333256 --minimum-latitude 4.961175132990646 --maximum-latitude 58.61512088772785 --minimum-depth 0 --maximum-depth 6000 -o /glade/work/emilanese/croc_input/dev_GLORY_PARALLEL_CESM-MOM6_NWA12_2023_cli/extract_forcings/raw_data -f east_unprocessed.2023-01-01_2023-01-10.nc &


copernicusmarine subset --dataset-id cmems_mod_glo_phy_my_0.083deg_P1D-m --variable so --variable thetao --variable uo --variable vo --variable zos --start-datetime 2023-01-11T00:00:00 --end-datetime 2023-01-20T23:59:59 --minimum-longitude -37.68856021347937 --maximum-longitude -36.093333333333256 --minimum-latitude 4.961175132990646 --maximum-latitude 58.61512088772785 --minimum-depth 0 --maximum-depth 6000 -o /glade/work/emilanese/croc_input/dev_GLORY_PARALLEL_CESM-MOM6_NWA12_2023_cli/extract_forcings/raw_data -f east_unprocessed.2023-01-11_2023-01-20.nc &


copernicusmarine subset --dataset-id cmems_mod_glo_phy_my_0.083deg_P1D-m --variable so --variable thetao --variable uo --variable vo --variable zos --start-datetime 2023-01-21T00:00:00 --end-datetime 2023-01-30T23:59:59 --minimum-longitude -37.68856021347937 --maximum-longitude -36.093333333333256 --minimum-latitude 4.961175132990646 --maximum-latitude 58.61512088772785 --minimum-depth 0 --maximum-depth 6000 -o /glade/work/emilanese/croc_input/dev_GLORY_PARALLEL_CESM-MOM6_NWA12_2023_cli/extract_forcings/raw_data -f east_unprocessed.2023-01-21_2023-01-30.nc &


copernicusmarine subset --dataset-id cmems_mod_glo_phy_my_0.083deg_P1D-m --variable so --variable thetao --variable uo --variable vo --variable zos --start-datetime 2023-01-31T00:00:00 --end-datetime 2023-02-01T23:59:59 --minimum-longitude -37.68856021347937 --maximum-longitude -36.093333333333256 --minimum-latitude 4.961175132990646 --maximum-latitude 58.61512088772785 --minimum-depth 0 --maximum-depth 6000 -o /glade/work/emilanese/croc_input/dev_GLORY_PARALLEL_CESM-MOM6_NWA12_2023_cli/extract_forcings/raw_data -f east_unprocessed.2023-01-31_2023-02-01.nc &


copernicusmarine subset --dataset-id cmems_mod_glo_phy_my_0.083deg_P1D-m --variable so --variable thetao --variable uo --variable vo --variable zos --start-datetime 2023-01-01T00:00:00 --end-datetime 2023-01-10T23:59:59 --minimum-longitude -98.98996846773888 --maximum-longitude -97.76 --minimum-latitude 4.961175132990646 --maximum-latitude 53.01719011051776 --minimum-depth 0 --maximum-depth 6000 -o /glade/work/emilanese/croc_input/dev_GLORY_PARALLEL_CESM-MOM6_NWA12_2023_cli/extract_forcings/raw_data -f west_unprocessed.2023-01-01_2023-01-10.nc &


copernicusmarine subset --dataset-id cmems_mod_glo_phy_my_0.083deg_P1D-m --variable so --variable thetao --variable uo --variable vo --variable zos --start-datetime 2023-01-11T00:00:00 --end-datetime 2023-01-20T23:59:59 --minimum-longitude -98.98996846773888 --maximum-longitude -97.76 --minimum-latitude 4.961175132990646 --maximum-latitude 53.01719011051776 --minimum-depth 0 --maximum-depth 6000 -o /glade/work/emilanese/croc_input/dev_GLORY_PARALLEL_CESM-MOM6_NWA12_2023_cli/extract_forcings/raw_data -f west_unprocessed.2023-01-11_2023-01-20.nc &


copernicusmarine subset --dataset-id cmems_mod_glo_phy_my_0.083deg_P1D-m --variable so --variable thetao --variable uo --variable vo --variable zos --start-datetime 2023-01-21T00:00:00 --end-datetime 2023-01-30T23:59:59 --minimum-longitude -98.98996846773888 --maximum-longitude -97.76 --minimum-latitude 4.961175132990646 --maximum-latitude 53.01719011051776 --minimum-depth 0 --maximum-depth 6000 -o /glade/work/emilanese/croc_input/dev_GLORY_PARALLEL_CESM-MOM6_NWA12_2023_cli/extract_forcings/raw_data -f west_unprocessed.2023-01-21_2023-01-30.nc &


copernicusmarine subset --dataset-id cmems_mod_glo_phy_my_0.083deg_P1D-m --variable so --variable thetao --variable uo --variable vo --variable zos --start-datetime 2023-01-31T00:00:00 --end-datetime 2023-02-01T23:59:59 --minimum-longitude -98.98996846773888 --maximum-longitude -97.76 --minimum-latitude 4.961175132990646 --maximum-latitude 53.01719011051776 --minimum-depth 0 --maximum-depth 6000 -o /glade/work/emilanese/croc_input/dev_GLORY_PARALLEL_CESM-MOM6_NWA12_2023_cli/extract_forcings/raw_data -f west_unprocessed.2023-01-31_2023-02-01.nc &

wait 
echo "All parallel calls to copernicusmarine have completed."
```

enrico-mi:
@manishvenu this is ready for view, the only thing I can't catch is why one of the tests hung

CrocoDash/raw_data_access/datasets/glorys.py
```diff
@@ -149,10 +154,12 @@ def get_glorys_data_script_for_cli(
         output_filename,
         variables=None,
         name=None,
+        **kwargs,
     ) -> None:
         """
         Script to run the GLORYS data query for the CLI
         """
+        last_pair = kwargs.pop("last_pair", False)
```
manishvenu: Can you elaborate a bit on why "last_pair" is used as the arg? I'm also guessing this could be an explcit default arg.

CrocoDash/extract_forcings/obc.py
```diff
@@ -201,92 +447,46 @@ def _regrid_boundary(
...
-                startdate=start_date,
-                repeat_year_forcing=False,
+    num_workers = min(available_cpus(), len(pairs))
+    pairs_per_workers = math.ceil(len(pairs) / num_workers)
+    regridded_files = []
+    logger.info("REGRID [%s]: ready for multi processes", boundary)
+    with ProcessPoolExecutor(max_workers=num_workers) as ex:
+        futures = [
+            ex.submit(
+                _regrid_per_process,
```
manishvenu, line 447: Could you explain a bit about why there is regrid_per_process and regrid_chunk?

manishvenu (commented):
Looks good, just a couple questions!

CrocoDash/extract_forcings/obc.py
enrico-mi: TL;DR: `_regrid_one_chunk` is there to isolate the rigridding of a single file (chunk) so that it can be called inside a loop or elsewhere; it's the minimal unit of code during regridding. I can integrate it inside `_regrid_per_process`.

The longer architectural explanation. When we get the boundary, each call (to RDA or copernicusmarine) requires only light data (date pairs, boundary name, output dir, etc.), so `_get_boundary` directly sends the list of chunks with their light data to the workers (l. 418): every processor takes the next available item in the list (with the light data) and calls the download method, once it's done it takes the next available item, etc.
For regridding it is slightly different. To process a chunk, we need to pass also the hgrid (l. 205), which is time-consuming if we were to open the hgrid file for every chunk. At the same time, if we were to open the file once and send a copy of the hgrid to each processor, it would add memory usage and serialization overhead. The solution is halfway: I build the list of chunks to process, split it based on the number of processors, and assign every processor its own sublist. Each processor opens the hgrid file once, and then sends it together with the light data to `_regrid_one_chunk()`. This minimizes hgrid serialization and memory usage, while the time required to open the hgrid once per processor is acceptable.
Having `_regrid_boundary()`, `_regrid_per_process()` and `_regrid_per_chunk()` makes the three layers more evident. (When we get the boundary, process and chunk levels overlap so we only have `_get_boundary()` and `_get_one_chunk()`.)

CrocoDash/raw_data_access/datasets/glorys.py
enrico-mi: this relates to reg mom 6 PR where I added that argument too (see [here](https://github.com/CROCODILE-CESM/regional-mom6/pull/72/changes#diff-2000ca73f82081d04bbd0ef1613e04a477ac2ffdc54bcb896b0272971eb31db5)). for the bash script to be executed in parallel, we need a `wait` statement at the end of the script, if I know that I'm passing the last pair of dates (i.e. I'm writing the instructions to download the last chunk), then I'm at the end of the script and can add the statement

CrocoDash/extract_forcings/obc.py
manishvenu: Ok great, thanks, this makes sense. Somewhat related, we just opened a PR on regional-mom6 the other day to take a slice of the hgrid instead of the entire thing.

CrocoDash/raw_data_access/datasets/glorys.py
manishvenu: Oh, okay yes yes. Needs to know if it's the last because writing to the same file. Can we potentiall rename this to last_request or something similar? last_pair was throwing me for a loop.

manishvenu:
I'll try to run this today and get a proper review in Enrico! Sorry for the wait

CrocoDash/extract_forcings/obc.py
```diff
@@ -161,32 +362,77 @@ def _get_boundary(
...
-            extra_args=extra_args,
+    with xr.open_dataset(hgrid_path) as hgrid:
+        latlon = Grid.get_bounding_boxes(hgrid)[boundary]
+
+    pairs = list(_make_date_pairs(start_date, end_date, get_step_days))
+
+    # Spread work across processes. If no chunking is prescribed it falls back
+    # to one processor. For get_glorys_data_script_for_cli(), generate CLI
+    # script that runs on multiple processes
+    if function_name == "get_glorys_data_script_for_cli":
```
manishvenu, line 379: Previously in `get_boundary` this function called utils.fetch_raw_chunk, which was very short. I see what was done here to make every possible function parallel and it makes total sense. The "script" methods are really tricky because they don't run in code. 

I think we should just ignore the script methods for now. It will output a script the user needs to run, and if they need to make it parallel, it's pretty simple to do. 

I don't like that we have to add so many lines of code just to pass last_boundary -> last_pair to handle the script method which also has the effect of change the regional-mom6 user API as well. 

I propose making this an issue we resolve later (potentially a new function in raw_data_access), given that this will still work for scripts.

CrocoDash/extract_forcings/obc.py
manishvenu: So when that comes through, I assume I can revisit and fold the two subfunctions together?

manishvenu (changes requested):
Hey Enrico, 

This looks really great, you made minimal changes, and this makes total sense to me. I left a couple comments on things I thought might simplify this PR. I'll run a case today to finalize my review, but let me know what you think.

CrocoDash/extract_forcings/obc.py
enrico-mi: Yea, I agree that the machinery is not elegant and I'll drop this approach.

One option for the future could be to leave the interaction with regional-mom6 untouched, and add a call that simply adds `&`s and `wait` to the script generated by regional-mom6.

CrocoDash/extract_forcings/obc.py
enrico-mi: I think so, can you link the PR? or if you have any conflict when you merge the two, I can look at them

CrocoDash/extract_forcings/obc.py
manishvenu: Yes, that would definetely work.

CrocoDash/extract_forcings/obc.py
manishvenu: https://github.com/COSIMA/regional-mom6/pull/315

manishvenu: and the respective PR on CrocoDash: https://github.com/CROCODILE-CESM/CrocoDash/pull/269, but I think we can handle this folding later, not a priority!

manishvenu:
parallel-fake-merge has the tentative merge with main.

CrocoDash/extract_forcings/obc.py
enrico-mi: done

manishvenu:
@enrico-mi Refactored the code to make obc.py seperate from mom6.py so we can use it for ww3. So had to make a few changes to merge main into this. I think everything got transferred over properly but might be worth a check!

CrocoDash/extract_forcings/obc.py
```diff
@@ -228,20 +426,17 @@ def _get_boundary(
     """Download all raw data for one boundary, chunked by get_step_days."""
     output_dir = Path(output_dir)
 
-    data_access_fn = utils.get_data_access_function(product_name, function_name)
+    pairs = list(_make_date_pairs(start_date, end_date, get_step_days))
```
alperaltuntas, line 429: The `dask.config.set(scheduler="synchronous")` block that #226 added here is gone, and
`_get_one_chunk` doesn't set it either. copernicusmarine calls `dask.compute()` inside
`to_netcdf()`, which then routes to the distributed scheduler and fails serializing
`botocore.client.S3`. The one in `_merge_boundary` is still there, so this looks like it got
lost in the merge rather than dropped on purpose. Can we set it inside `_get_one_chunk` so
each worker gets it?

CrocoDash/extract_forcings/obc.py
```diff
@@ -208,6 +213,199 @@ def _validate_coverage(
...
+                "Delete it and re-run."
+            )
+        logger.info(f"Regridded file {dated_output.name} already exists. Skipping.")
+        return dated_output, regridders
+
+    logger.info("PROC [%d] - REGRID [%s]: Validating coverage", proc_id, boundary)
+    parse_raw_dates = lambda f, boundary=boundary: _parse_raw_filename_dates(
+        f, boundary
+    )
+    chunk_raw_files = _validate_coverage(
```
alperaltuntas, line 351: `_files_within_range` keeps only files entirely inside the window, and `_validate_coverage`
then wants exact endpoints, so a raw file wider than one regrid chunk gets dropped here. A
generated case is fine because `configurations.py` writes `get_step_days` and
`regrid_step_days` equal, but lowering `regrid_step_days` alone fails with `No files for
[east]`, which is misleading since the raw data is there. The module docstring still says GET
and REGRID chunks are fully independent. Can we make the check overlap-based, or leave it to
the whole-range validation in `process_obc_conditions`?

CrocoDash/extract_forcings/obc.py
```diff
@@ -269,76 +484,82 @@ def _regrid_boundary(
...
-    ):
-        start_str = chunk_start.strftime("%Y-%m-%d")
-        end_str = chunk_end.strftime("%Y-%m-%d")
-        dated_output = (
-            output_folder / f"forcing_obc_segment_{seg_id:03d}_{start_str}_{end_str}.nc"
+    pairs = list(_make_date_pairs(start_date, end_date, regrid_step_days))
+    if not pairs:
+        return []
+
+    num_workers = max(1, min(available_cpus(), len(pairs)))
```
alperaltuntas, line 509: This branch is never taken in CI. The only test that reaches `_regrid_boundary` for real is
marked slow, and `test_process_obc_conditions_uses_tmask_bbox_for_get` monkeypatches both
phase functions away, so every run takes `num_workers == 1`. That's what codecov/patch is
flagging too. Can we add a small test that forces `num_workers > 1` over a couple of tiny
chunks, so the worker slicing and the `os.replace` stay covered?

CrocoDash/extract_forcings/obc.py
```diff
@@ -253,6 +448,26 @@ def _get_boundary(
                 variables=variables,
                 extra_args=extra_args,
             )
+    else:
+        num_workers = min(available_cpus(), len(pairs))
```
alperaltuntas, line 452: No `max(1, ...)` here, unlike line 509. If `pairs` is empty this gives
`ProcessPoolExecutor(max_workers=0)`, which raises. Same expression otherwise, so this
looks like an oversight. please add the guard.

CrocoDash/extract_forcings/obc.py
```diff
@@ -253,6 +448,26 @@ def _get_boundary(
...
+        with ProcessPoolExecutor(max_workers=num_workers) as ex:
```
alperaltuntas, line 453: Both pools take the platform default start method. That's `fork` on linux today, but
3.14 switches linux to `forkserver`, and `spawn`/`forkserver` re-import a bare
`ProductRegistry` in the child, so any product registered at runtime is lost. This is
exactly what broke when I registered a local GLORYS product for prestaging. Since this
PR is where the process pools land, can we pass
`mp_context=multiprocessing.get_context("fork")` on both this one and line 541?

CrocoDash/raw_data_access/datasets/glorys.py
```diff
@@ -92,14 +92,19 @@ def get_glorys_data_from_rda(
         ds_in_files = sorted(ds_in_files)
 
         ds = xr.open_mfdataset(
-            ds_in_files, decode_times=False, engine="h5netcdf", parallel=True
+            ds_in_files,
+            decode_times=False,
+            engine="h5netcdf",
+            parallel=False,
```
alperaltuntas, line 98: This was `parallel=True` on main. With `get_step_days` defaulting to the whole period,
the chunked path is never taken, so the default case now opens the per-day files without
dask parallelism and probably ends up slower than main rather than faster. Was flipping
this deliberate, or a leftover from testing the chunked path?

CrocoDash/extract_forcings/obc.py
```diff
@@ -208,6 +213,199 @@ def _validate_coverage(
...
+    logger.info(
+        "PROC [%d] - REGRID [%s]: Regridding done for %s",
+        proc_id,
+        boundary,
+        dated_output,
+    )
+    return dated_output, regridders
+
+
+def available_cpus():
```
alperaltuntas, line 400: `sched_getaffinity` returns 128 on a derecho login node, where the user slice is
actually capped at 10 GB and the arbiter gives ~8 cores, so this over-reports badly
there. In a batch job it is correct, since PBS cpusets bind the allocation, so a flat
cap would just waste a 128-core node. The distinction we have is `PBS_JOBID`: unset
means login node.

Can we use the full affinity when `PBS_JOBID` is set, and cap conservatively otherwise,
with an env var to override either way? Benchmarking the per-day GLORYS reads I got a
worker OOM-killed at 8 processes on a login node with a notebook holding 4 to 5 GB, and
peak RSS grows about 0.14 GB per worker.

Granted, we don't expect users to run production extractions on login nodes. But it'd
still be good to keep these limits in mind, because `process_forcings()` is a notebook
call and people do run it interactively while setting a case up, which on derecho lands
on a login node or a small JupyterHub allocation. When it goes wrong it surfaces as a
bare `BrokenProcessPool` with no indication that memory was the cause, so it reads as a
CrocoDash bug and comes back to us as an issue. We also have ~80 people doing exactly
this at once in the workshop practical.

Note the PR description still says workers are capped at 12, which 4ff52d61 removed.

Separately, the fallback returns 12 on any platform without `sched_getaffinity`, which
claims 12 cpus on a 4-core laptop. `os.cpu_count()` is probably the better fallback.

alperaltuntas (changes requested):
Thanks a lot, @enrico-mi! This is all great. @manishvenu, below are some final issues that need resolving, so please address them. In addition to the line comments below, there is one more thing that isn't anchored to a line, since the file isn't in this PR:

`configurations.py` sets both `get_step_days` and `regrid_step_days` to the entire
modeling period:

    step = (end_dt - start_dt).days + 1
    self.set_output_param("get_step_days", step)
    self.set_output_param("regrid_step_days", step)

So a generated case always gets `len(pairs) == 1`, and both process pools fall through
to the serial path. As merged, the parallelism in this PR never actually runs unless
someone hand-edits `config.json`.

That one default is behind three of my line comments: it's why `_regrid_boundary`'s
multi-worker branch is uncovered in CI, why the `_files_within_range` mismatch doesn't
show up on a generated case, and why flipping `parallel=False` in `glorys.py` leaves the
default path with no parallelism at either level. Fixing it in `configurations.py`
addresses all three at the source, rather than three times over.

Can we default these to something smaller than the whole period, e.g., 30 days? That
would need to come with, or after, a fix for the `_files_within_range` issue, since
unequal get and regrid steps is exactly what currently fails.

CrocoDash/forcing/obc.py
```diff
@@ -228,20 +426,17 @@ def _get_boundary(
     """Download all raw data for one boundary, chunked by get_step_days."""
     output_dir = Path(output_dir)
 
-    data_access_fn = utils.get_data_access_function(product_name, function_name)
+    pairs = list(_make_date_pairs(start_date, end_date, get_step_days))
```
manishvenu: Done.

CrocoDash/forcing/obc.py
```diff
@@ -208,6 +213,199 @@ def _validate_coverage(
...
+                "Delete it and re-run."
+            )
+        logger.info(f"Regridded file {dated_output.name} already exists. Skipping.")
+        return dated_output, regridders
+
+    logger.info("PROC [%d] - REGRID [%s]: Validating coverage", proc_id, boundary)
+    parse_raw_dates = lambda f, boundary=boundary: _parse_raw_filename_dates(
+        f, boundary
+    )
+    chunk_raw_files = _validate_coverage(
```
manishvenu: Done to overlapped based.

CrocoDash/forcing/obc.py
```diff
@@ -269,76 +484,82 @@ def _regrid_boundary(
...
-    ):
-        start_str = chunk_start.strftime("%Y-%m-%d")
-        end_str = chunk_end.strftime("%Y-%m-%d")
-        dated_output = (
-            output_folder / f"forcing_obc_segment_{seg_id:03d}_{start_str}_{end_str}.nc"
+    pairs = list(_make_date_pairs(start_date, end_date, regrid_step_days))
+    if not pairs:
+        return []
+
+    num_workers = max(1, min(available_cpus(), len(pairs)))
```
manishvenu: Yes

CrocoDash/forcing/obc.py
```diff
@@ -253,6 +448,26 @@ def _get_boundary(
                 variables=variables,
                 extra_args=extra_args,
             )
+    else:
+        num_workers = min(available_cpus(), len(pairs))
```
manishvenu: done

CrocoDash/forcing/obc.py
```diff
@@ -253,6 +448,26 @@ def _get_boundary(
...
+        with ProcessPoolExecutor(max_workers=num_workers) as ex:
```
manishvenu: done

CrocoDash/raw_data_access/datasets/glorys.py
manishvenu: I will test this as it is! I believe it was a long timeago changed because of hanging issues, but we're a while from that change.

CrocoDash/forcing/obc.py
```diff
@@ -208,6 +213,199 @@ def _validate_coverage(
...
+    logger.info(
+        "PROC [%d] - REGRID [%s]: Regridding done for %s",
+        proc_id,
+        boundary,
+        dated_output,
+    )
+    return dated_output, regridders
+
+
+def available_cpus():
```
manishvenu: Implemented this exact design.

manishvenu:
@alperaltuntas I fixed the review comments, I'll run a test to see if it's fully working.

CrocoDash/raw_data_access/datasets/glorys.py
manishvenu: It's working now on the login node  with parallel either one, but I'm not sure if we'll run into errors either way, I'll switch it back to true for now.

manishvenu:
Test works!

manishvenu:
Tests may need to be updated depending on CI

manishvenu:
I don't necessarily understand this last commit as much, but I believe it's a pretty minimal change and passes tests now.

alperaltuntas (approved):

PR #272, opened 2026-08-18
Fix custom compset creation for multi-modifier physics options

manishvenu:
**Descrip:**

`Case()` silently fails for any compset long name with 2+ physics modifiers on one component, e.g. `MOM6%REGIONAL%MARBL-BIO`. Only custom compsets are affected; anything selected via a registered CIME alias (e.g. `CR1850MARBL_JRA_GLOFAS`) goes through.

**Changes:**

1. Root cause: `_configure_custom_compset`'s per-component loop assigns each component's full `%`-joined modifier string to its `COMP_{class}_OPTION` cvar in one step. Each assignment fires `compset_vars.py`'s `update_compset_lname` observer, which immediately rebuilds and Z3-validates the *complete* `COMPSET_LNAME` from whatever `COMP_*_OPTION` values are set so far in the loop — including components not yet reached (ROF's `%GLOFAS`, assigned after OCN). It validates a genuinely invalid intermediate string, raises `ConstraintViolation`, and `Case.__init__`'s bare `except Exception: print(...); return` swallows it silently.

2. Fix: assign only the *first* modifier per component during this stage. That matches the surviving comment a few lines below, which already describes this design — the full multi-modifier `COMPSET_LNAME` gets force-assigned once the whole Options stage completes. A later refactor (`67cd3449`, "Refactor OBC forcing pipeline") changed the loop to join all modifiers instead of taking just the first, breaking this without updating the comment.


manishvenu:
See #279

